PR #3, opened 2015-02-05
Warn on bad CSV filenames and remove them

tie-rack:
Bad files in an upload should only produce warnings and be removed from the context, not stop the processing.

[Pivotal story](https://www.pivotaltracker.com/story/show/87562548)


tank157:
LGTM

PR #4, opened 2015-02-06
RabbitMQ

tie-rack:
Pivotal stories: [86533368](https://www.pivotaltracker.com/story/show/86533368) and [86159336](https://www.pivotaltracker.com/story/show/86159336)

### DIFF
--- a/src/vip/data_processor/queue.clj
+++ b/src/vip/data_processor/queue.clj
@@ -9,7 +9,6 @@
 
 (def rabbit-connection (atom nil))
 (def rabbit-channel (atom nil))
-(def qa-engine-exchange "qa-engine")
 
 (defn initialize []
   (let [max-retries 5]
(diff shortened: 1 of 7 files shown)


src/vip/data_processor/queue.clj
```diff
@@ -0,0 +1,45 @@
...
+            [langohr.core :as rmq]
+            [langohr.channel :as lch]
+            [langohr.exchange :as le]
+            [langohr.queue :as lq]
+            [langohr.basic :as lb]
+            [turbovote.resource-config :refer [config]]))
+
+(def rabbit-connection (atom nil))
+(def rabbit-channel (atom nil))
+(def qa-engine-exchange "qa-engine")
```
tank157: Should we get the exchange name from config?

tie-rack: Will do. Commit coming.

tank157: Looks good otherwise

PR #5, opened 2015-02-11
Postgres

tie-rack:
Add Postgres to the QA Engine.

Pivotal stories: [86159554](https://www.pivotaltracker.com/story/show/86159554) and [86159782](https://www.pivotaltracker.com/story/show/86159782)

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,18 @@ $ lein trampoline run
 ### Docker
 
 Configuration in Docker is set with environment variables, so you'll
-need to set them when you run the container. See the variable names in
-`resources/config.edn`.
+need to set them when you run the container. The following variables
+need values:
+
+* `VIP_DP_AWS_ACCESS_KEY`
+* `VIP_DP_AWS_SECRET_KEY`
+* `VIP_DP_S3_BUCKET`
+* `VIP_DP_SQS_REGION`
+* `VIP_DP_SQS_QUEUE`
+* `VIP_DP_SQS_FAIL_QUEUE`
+* `VIP_DP_RABBITMQ_EXCHANGE`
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -22,8 +22,18 @@ $ lein trampoline run
...
+
+* `VIP_DP_AWS_ACCESS_KEY`
+* `VIP_DP_AWS_SECRET_KEY`
+* `VIP_DP_S3_BUCKET`
+* `VIP_DP_SQS_REGION`
+* `VIP_DP_SQS_QUEUE`
+* `VIP_DP_SQS_FAIL_QUEUE`
+* `VIP_DP_RABBITMQ_EXCHANGE`
+* `POSTGRES_USER`
+* `POSTGRES_PASSWORD`
```
cap10morgan: Is there any way to supply sane defaults for these in dev (w/ docker-compose) while still allowing them to be overidden when necessary? Do they need to be overridden in dev?

tie-rack: There's certainly not sane defaults for the AWS credentials. And I'd be wary of supplying other defaults that could accidentally get used in production (since they'd probably be wrong).

cap10morgan: For AWS creds: Yes, you can't default those. Could we use `AWS_ACCESS_KEY` and `AWS_SECRET_KEY` similarly to other tools?

For the others: Why would they accidentally get used in production if the defaults were just in `docker-compose.yml`?

tie-rack: It sounds like maybe you want to open an issue ;)

This section of the README is merely making more explicit a situation that already exists and this conversation probably belongs outside of this PR.

cap10morgan: Sure, will do.

PR #8, opened 2015-02-18
Sqlite

tie-rack:
[Pivotal story](https://www.pivotaltracker.com/story/show/87534884)

The `vip.data-processor.db.sqlite/entity` macro is not used here, but is about to be. It's simply Korma's `defentity` macro minus the `def` part. We don't want to be defining anything, since we'll be making a bunch of temporary databases in parallel. Using the entity value itself off a key on the processing context will work just as well.

For example:
```clojure
(korma/select (get-in ctx [:db :entities :elections]))
```

### DIFF
--- a/src/vip/data_processor/db/sqlite.clj
+++ b/src/vip/data_processor/db/sqlite.clj
@@ -1,0 +1,24 @@
+(ns vip.data-processor.db.sqlite
+  (:require [joplin.core :as j]
+            [joplin.jdbc.database]
+            [korma.db :as db]
+            [korma.core :as korma])
+  (:import [java.nio.file Files]
+           [java.nio.file.attribute FileAttribute]))
+
+(defn create-temp-db-file [prefix]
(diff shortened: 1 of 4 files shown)


src/vip/data_processor/db/sqlite.clj
```diff
@@ -0,0 +1,24 @@
+(ns vip.data-processor.db.sqlite
+  (:require [joplin.core :as j]
+            [joplin.jdbc.database]
```
tank157: Is there a reason to require this?

tie-rack: Joplin [demands it](https://github.com/juxt/joplin#calling-joplin-from-your-code)

> Calling joplin from your code requires that you have loaded the joplin.core namespace and also the namespaces of the database plugins you intend to use.

tank157:
Ok, LGTM. I guess merge this in quick because it looks like you're re-commiting this work in other branches that depend on it?

tie-rack:
:) Thanks, Troy.

PR #10, opened 2015-02-19
Load source

bwreid:
Loads source.txt file.

[Pivotal Card](https://www.pivotaltracker.com/story/show/87563166) & [Pivotal Card](https://www.pivotaltracker.com/story/show/87562722)

Assigned @tank157.

### DIFF
--- a/src/vip/data_processor/validation/csv.clj
+++ b/src/vip/data_processor/validation/csv.clj
@@ -1,5 +1,8 @@
 (ns vip.data-processor.validation.csv
-  (:require [clojure.set :as set]))
+  (:require [clojure.data.csv :as csv]
+            [clojure.java.io :as io]
+            [clojure.set :as set]
+            [korma.core :as korma]))
 
 (def csv-filenames
   #{"ballot.txt"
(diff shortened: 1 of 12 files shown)


src/vip/data_processor/validation/csv.clj
```diff
@@ -51,3 +54,39 @@
...
+        (korma/insert elections-table (korma/values coerced-contents))
+        ctx)
+      (assoc-in ctx [:errors :load-elections] "election.txt missing"))))
+
+(defn load-sources [ctx]
+  (let [files (:input ctx)
+        source-file (first (filter #(= "source.txt" (.getName %)) files))]
+    (if source-file
+      (let [sources-table (get-in ctx [:tables :sources])
+            contents (read-csv-with-headers source-file)]
```
tank157: Not that it's in the current scope of the work, but at some phase there will be validation of the headers of the files, I assume based off of the way things work in quick_feed/Metis. So you might want to consider how that will go into the flow sooner rather than later so you maybe don't have to re-write things for 20+ file types later. Just a suggestion.

tank157:
LGTM

src/vip/data_processor/validation/csv.clj
tie-rack: Validating the headers can be a separate validation fn. We don't need to make it a part of the loading functions.

tank157: Yeah, Wesley and I are discussing it and it does seem like it'll be an earlier phase issue. But something else to consider right now, is that extra columns are likely to happen, should be ignored, and the loading you have currently won't handle that. So at some point you'll probably want to be able to inject the column specs in to use them for like a select-keys call.

tie-rack: I'll try something now that we have a couple of these. Refactor coming!

PR #11, opened 2015-02-19
Load states

bwreid:
Loads the state.txt file. If it doesn't exist, adds a warning.

[Pivotal Card](https://www.pivotaltracker.com/story/show/87563418)

Assigned @tie-rack.


tie-rack:
Awesome. Merging.

PR #12, opened 2015-02-19
csv-loaders and missing-file fns

tie-rack:
Now that we've built a few, let's make it a nicer experience:

For example, a processing function that loads a CSV:
```clojure
(def load-elections (csv-loader "election.txt" :elections
                                (booleanize "election_day_registration")
                                (booleanize "statewide")))
```

This will load the data in "elections.txt" into the `:elections` table, applying those two `booleanize` functions to each row in the CSV. It will also ignore any missing columns (instead of, as before, blowing up).

Also, a processing function that adds an error if a file is missing:
```clojure
(error-on-missing-file "election.txt")
```

This functionality was moved from the loading functions because it's a different job. There is also a similar processing function generator called `warn-on-missing-file`.

### DIFF
--- a/src/vip/data_processor/validation/csv.clj
+++ b/src/vip/data_processor/validation/csv.clj
@@ -91,7 +91,8 @@
             contents (read-csv-with-headers file-to-load)
             transforms (apply comp select-columns row-transform-fns)
             transformed-contents (map transforms contents)]
-        (korma/insert sql-table (korma/values transformed-contents))))
+        (doseq [row transformed-contents]
+          (korma/insert sql-table (korma/values row)))))
     ctx))
 
 (defn add-report-on-missing-file-fn
--- a/src/vip/data_processor/validation/csv.clj
+++ b/src/vip/data_processor/validation/csv.clj
@@ -68,35 +68,53 @@
                   (if (= "yes" (row field))
                     1 0))))
 
-(defn load-elections [ctx]
-  (let [files (:input ctx)
-        election-file (first (filter #(= "election.txt" (.getName %)) files))]
-    (if election-file
-      (let [elections-table (get-in ctx [:tables :elections])
-            contents (read-csv-with-headers election-file)
-            coerced-contents (->> contents
-                                  (map (booleanize "election_day_registration"))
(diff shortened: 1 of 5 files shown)


src/vip/data_processor/validation/transforms.clj
```diff
@@ -27,6 +27,9 @@
 
 (def csv-validations
   [csv/remove-bad-filenames
+   (csv/error-on-missing-file "election.txt")
+   (csv/error-on-missing-file "source.txt")
+   (csv/warn-on-missing-file "state.txt")
```
bwreid: At some point I can see us wanting to move all of these calls to either a separate function and/or a list (eg. `(csv/error-on-missing-files ["election.txt" "source.txt" ...]`.

For now, this is great though.

src/vip/data_processor/validation/csv.clj
```diff
@@ -68,35 +69,54 @@
...
+(def ^{:doc "Generates a validation function that adds an error when
+  the given filename is missing from the input"}
+  error-on-missing-file
+  (add-report-on-missing-file-fn :errors))
+
+(def load-elections (csv-loader "election.txt" :elections
+                                (booleanize "election_day_registration")
+                                (booleanize "statewide")))
+(def load-sources (csv-loader "source.txt" :sources))
+(def load-states (csv-loader "state.txt" :states))
```
bwreid: Is there a reason you want both the name of a file and a key as opposed to something (perhaps harmfully) clever that uses the filename to create the key / vice versa? It just seems like we're hardcoding two things.

tie-rack: I'd advise against writing a function that generates one from the other, especially as pluralization rules are going to come into play. The _domain_ of such a function (in our case) is small enough that stating the values like this (or in a map from one to the other) is not so harmful. (I think.)

src/vip/data_processor/validation/transforms.clj
tie-rack: That's true. That _will_ come in handy.

bwreid:
Sweet. Looks good to me.

PR #13, opened 2015-02-24
Load officials and administrations

tie-rack:
Pivotal stories: [87558696](https://www.pivotaltracker.com/story/show/87558696) and [87561576](https://www.pivotaltracker.com/story/show/87561576)


bwreid:
Tests added! Looks good to me.

PR #14, opened 2015-02-24
Buildkite build script

tie-rack:
This is enough to get us started on getting green checks or red exes when we make PRs.

`docker build` runs the tests.

### DIFF
--- a/script/build
+++ b/script/build
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -exo pipefail
+
+IMAGE_TAG=${BUILDKITE_TAG:-"$BUILDKITE_BRANCH-$(git rev-parse --short $BUILDKITE_COMMIT)"}
+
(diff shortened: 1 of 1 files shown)


cap10morgan:
I'd like us to hold ourselves to a standard of keeping these usable with and without BuildKite. So a) Do you agree with that approach? and b) Do you want to make the case to get this in place now and create a card for doing that later, or does it seem simple enough to do in here?

Obviously you could use it locally by setting all of the BUILDKITE_* env vars, but the approach I've taken in other scripts was to check for the top-level var (so IMAGE_TAG in this case) being already set (or provided as an arg to the script) and using that when it is.

tie-rack:
Perhaps just checking for `BUILDKITE` and if so, set `IMAGE_TAG` like that, otherwise, use the branch and short SHA of `HEAD`?

cap10morgan:
Yep, sounds even better.

cap10morgan:
LGTM

PR #15, opened 2015-02-24
Load localities and precincts

bwreid:
Assigned @tie-rack.

[Pivotal Card This](https://www.pivotaltracker.com/story/show/87561212) & [Pivotal Card That](https://www.pivotaltracker.com/story/show/87561640)


tie-rack:
Looks great!

(Buildkite failure due to not having the `buildkite` branch baked in. So it's okay.)

PR #17, opened 2015-02-26
Load polling locations and street segments

tie-rack:
SQL loading of polling_location.txt and street_segment.txt files.

Pivotal stories: [87561668](https://www.pivotaltracker.com/story/show/87561668) and [87563448](https://www.pivotaltracker.com/story/show/87563448)


bwreid:
This looks good, let's merge it!   :D

PR #18, opened 2015-03-03
[CS] Lazily process CSV instead of eagerly reading

tie-rack:
Large CSVs may be too large to fit in memory. So read them piecemeal.


tank157:
LGTM

PR #20, opened 2015-03-04
[CS] Insert multiple rows during csv import

tie-rack:
`chunk-rows` breaks up rows for import into as big chunks as possible for a SQLite insert. We can be sure we can import all rows for an import at once, since SQLite can only support 999 statement parameters at a time. This breaks them down into groups such that no group has more than 999 values to insert.

This *dramatically* speeds up imports of large files, especially street_segment.txt.

[Pivotal story](https://www.pivotaltracker.com/story/show/89618646)


cap10morgan:
LGAFTM (Looks Good And Fast To Me)

PR #23, opened 2015-03-10
Load precinct districts and sites

7maples:
Add CSV loading capability for:
* precinct_electoral_district.txt
* precinct_early_vote_site.txt

Pivotal cards: [89140714](https://www.pivotaltracker.com/story/show/89140714) and [89140516](https://www.pivotaltracker.com/story/show/89140516)

The data processor can now upload this data into individual tables and make changes (i.e. boolean values) where appropriate.

### DIFF
--- a/resources/processing-migrations/20150310-3-load-precinct-electoral-districts.up.sql
+++ b/resources/processing-migrations/20150310-3-load-precinct-electoral-districts.up.sql
@@ -1,0 +1,3 @@
+CREATE TABLE precinct_electoral_districts (id INTEGER PRIMARY KEY,
+                                           precinct_id INTEGER,
+                                           electoral_district_id INTEGER);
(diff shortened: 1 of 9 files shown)


resources/processing-migrations/20150310-3-load-precinct-electoral-districts.up.sql
```diff
@@ -0,0 +1,3 @@
+CREATE TABLE precinct_electoral_districts (id INTEGER PRIMARY KEY,
```
tie-rack: precinct-electoral-district.txt doesn't have an "id" column, so this table doesn't need on either.

resources/processing-migrations/20150310-4-load-precinct-early-vote-sites.up.sql
```diff
@@ -0,0 +1,3 @@
+CREATE TABLE precinct_early_vote_sites (id INTEGER PRIMARY KEY,
```
tie-rack: precinct_early_vote_site.txt doesn't have an "id" column, so this table doesn't need on either.

PR #24, opened 2015-03-11
File set validations

tie-rack:
Pivotal story: [89464334](https://www.pivotaltracker.com/story/show/89464334)

Add validations to check for file dependencies for any files present in an upload. For example, if "state_early_vote_site.txt" is present, validates that "state.txt" and "early_vote_site.txt" are also present.

Also removes unnecessary missing file warnings now that we have real dependency validations.


bwreid:
Looks incredible to me. :+1:

PR #25, opened 2015-03-12
Validate CSV header row

tie-rack:
Validate that every CSV has a header row, adding a "critical" error for each file missing one.

Pivotal story [90224778](https://www.pivotaltracker.com/story/show/90224778)


test/vip/data_processor/test_helpers.clj
```diff
@@ -0,0 +1,21 @@
+(ns vip.data-processor.test-helpers
+  (:require [clojure.test :refer :all]))
+
+(def problem-types [:warnings :errors :critical :fatal])
+
+(defn assert-no-problems
+  "Test that there are no errors of any level for the key-path below
```
7maples: Doc strings make me so happy

7maples:
Is there a Pivotal card for this? Otherwise LGTM

tie-rack:
Pivotal card is coming. The validation stories are going to be redone this afternoon. I'll link it here when it's written.

src/vip/data_processor/validation/csv.clj
```diff
@@ -59,7 +59,7 @@
   (let [raw-rows (csv/read-csv file-handle)
         headers (first raw-rows)
         rows (rest raw-rows)]
-    (map (partial zipmap headers) rows)))
+    [headers (map (partial zipmap headers) rows)]))
```
bwreid: Why not return something more explicit? `{:headers headers :rows (map (partial ...` You'd be able to destructure it in the same way below but it might be a bit more readable. Just my opinion though.

tie-rack: Good call. Update coming.

tie-rack:
Pivotal story added to the PR text.

PR #28, opened 2015-03-18
Validate CSV values and their formats

tie-rack:
Pivotal stories: [90227234](https://www.pivotaltracker.com/story/show/90227234), [90227616](https://www.pivotaltracker.com/story/show/90227616)

Validations for each row of CSV imports that cover required columns having values and that each value is of the correct format (e.g, starting with "http://" or is "yes"/"no" or is formatted like "(XXX) XXX-XXXX").

Includes updates to "full-good-run" example files that were less than "fully good".

The way CSVs are specified has been changed wholesale. Instead of a bunch of function calls to set up each importer, there is now a list of maps that describe each CSV and its columns. That list is read and turned in to the importers and validators that get run. These descriptions should be very readable and easily editable by people less familiar with Clojure code, but familiar with the CSVs themselves.


src/vip/data_processor/validation/csv.clj
```diff
@@ -3,47 +3,294 @@
...
 
 (defn file-name [file]
   (.getName file))
 
 (defn good-filename? [file]
-  (->> file
-       file-name
-       (contains? csv-filenames)))
+  (let [filename (file-name file)]
+    (contains? csv-filenames filename)))
```
7maples: I really like the explicit `contains?`, it's more readable this way

src/vip/data_processor/validation/csv.clj
```diff
@@ -3,47 +3,294 @@
...
-    "precinct_split_electoral_district.txt"
-    "precinct_split_polling_location.txt"
-    "referendum.txt"
-    "referendum_ballot_response.txt"
-    "source.txt"
-    "state.txt"
-    "state_early_vote_site.txt"
-    "street_segment.txt"})
+(defn boolean-value [x]
+  (if (re-find #"\A(?i:yes)\z" x) 1 0))
```
7maples: Do we want to make this more flexible so it can handle whitespace?

tie-rack: Well, the `yes-no` format validator is more strict than even this. The validations spec is pretty unforgiving. If there's whitespace, you're already going to get an error about the value.

test/vip/data_processor/validation/csv_test.clj
```diff
@@ -88,25 +34,80 @@
...
+          (let [result-ctx (format-rule ctx {} line-number)]
+            (is (= ctx result-ctx))))
+        (testing "if it is there"
+          (testing "it matches the format, everything's okay"
+            (let [result-ctx (format-rule ctx {column "1234"} line-number)]
+              (is (= ctx result-ctx))))
+          (testing "it doesn't match the format, you get an error"
+            (let [result-ctx (format-rule ctx {column "asdf"} line-number)]
+              (is (get-in result-ctx [:errors filename line-number column])))))))
+    (testing "check that is a list of options"
```
7maples: at first read this was awkward - maybe "list of options check" or "a check that is a list of options"?

src/vip/data_processor/validation/csv.clj
7maples: great!

7maples:
LGTM

PR #29, opened 2015-03-18
Validate number of values in each CSV row

tie-rack:
Pivotal story: [90226200](https://www.pivotaltracker.com/story/show/90226200)

Validate that the number of data columns in each file match the number of column headers for each row. Adds a critical error for each bad row.


bwreid:
I like this. LGTM.

PR #30, opened 2015-03-18
[WR] check for .txt file extensions when loading files

bwreid:
[Pivotal](https://www.pivotaltracker.com/story/show/90225200)

This PR provides a separate, clearer error when a file is not of the '.txt' file type.


src/vip/data_processor/validation/csv.clj
```diff
@@ -394,7 +396,11 @@
                   transformed-contents (map transforms contents)]
               (sqlite/bulk-import transformed-contents sql-table)
               ctx)))))
-    ctx))
+      ctx)))
+
+(defn load-csvs [csv-specs]
+  (fn [ctx]
+    (reduce load-csv ctx csv-specs)))
```
tie-rack: Is `load-csvs` in here twice now?

src/vip/data_processor/validation/csv.clj
```diff
@@ -371,8 +371,10 @@
        (map create-translation-fn)))
 
 (defn load-csv [ctx {:keys [filename table columns] :as csv-spec}]
-  (if-let [file-to-load (find-input-file ctx filename)]
-    (with-open [in-file (io/reader file-to-load)]
+  (if (nil? (re-seq #".txt" filename))
```
tie-rack: This might be slightly more "correct" (for some definition of correct) as `(if-not (.endsWith filename ".txt")`. The current regular expression would match file names like "eightxten.whatever".

src/vip/data_processor/validation/csv.clj
```diff
@@ -371,8 +371,10 @@
...
+    (update-in ctx [:critical filename] conj "File is not a .txt file.")
```
tie-rack: Also, the way these maps are being used as changed somewhat. Errors don't share lists of strings, but are maps all the way down to the individual validation. So this should become something like:

```clojure
  (assoc-in ctx [:critical filename "File extension"] "File is not a .txt file.")
```

test/vip/data_processor/validation/csv_test.clj
```diff
@@ -33,6 +33,16 @@
       (is (not (contains? (:errors out-ctx) "source.txt"))))))
 
 (deftest csv-loader-test
+  (testing "check for .txt file extension"
+    (let [only-election-official-spec (filter #(= "election_official.txt" (:filename %))
+                                              csv-specs)
+          updated-spec (list (assoc (first only-election-official-spec)
+                                    :filename "election_official.csv"))
```
tie-rack: So this is the crux of the problem with this story. There's no spec for a file called "election_official.csv", so a file with that name would never get loaded. In fact, before we even get to loading files in the real pipeline, this file would have already been marked as having an bad name and wouldn't make it to the `load-csvs` step.

In practice, this warning will never surface.

tie-rack:
Closing. This may turn into just a new error message for "bad filename" error.

PR #32, opened 2015-03-20
Duplicate records

tie-rack:
Pivotal story: [90228812](https://www.pivotaltracker.com/story/show/90228812)

Validate that CSV uploads do not contain duplicate records, according to a scan of all columns except ID, adding a warning for any such row found.

Also, removes duplicate rows from the full-good-run test data so the full-good-run produces no warnings.


src/vip/data_processor/validation/db.clj
```diff
@@ -28,3 +29,52 @@
...
+(defn validate-no-duplicated-rows-in-table [ctx {:keys [filename table]}]
+  (let [table (get-in ctx [:tables table])
+        potential-dupes (find-potential-dupes table)]
+    (if (seq potential-dupes)
+      (assoc-in ctx [:warnings filename :duplicated-rows] potential-dupes)
+      ctx)))
+
+(defn validate-no-duplicated-rows [csv-specs]
+  (fn [ctx]
+    (reduce validate-no-duplicated-rows-in-table ctx csv-specs)))
```
bwreid: Are all of the earlier functions that lead up to this fn going to be reused? If not, should we potentially start splitting these out into separate namespaces? I can see a generic 'db validation' file getting large.

tie-rack: That's a good point. I'll move them into something like `vip.data-processor.validation.db.duplicate-rows`.

bwreid:
The separate name-spacing makes me so happy (and makes it all so much more readable IMO). LGTM.

PR #33, opened 2015-03-24
Validate referential integrity

tie-rack:
Pivotal story: [90228318](https://www.pivotaltracker.com/story/show/90228318)

Validate that references to IDs in other files exist (e.g., that an `election_id` in context.txt matches an `id` in election.txt). Validates `jurisdiction_id` fields as well (which must match an `id` in one of state.txt, locality.txt, precinct.txt, precinct_split.txt, or electoral_districts.txt).

Adds errors for each unmatched reference.


7maples:
LGTM

src/vip/data_processor/db/sqlite.clj
```diff
@@ -13,6 +13,38 @@
...
+   :precinct-split-electoral-districts (entity :precinct_split_electoral_districts (korma/database db))
+   :early-vote-sites (entity :early_vote_sites (korma/database db))
+   :precinct-polling-locations (entity :precinct_polling_locations (korma/database db))
+   :polling-locations (entity :polling_locations (korma/database db))
+   :street-segments (entity :street_segments (korma/database db))
+   :electoral-districts (entity :electoral_districts (korma/database db))
+   :contests (entity :contests (korma/database db))
+   :ballots (entity :ballots (korma/database db))
+   :ballot-responses (entity :ballot_responses (korma/database db))
+   :referendums (entity :referendums (korma/database db))
```
jungshadow: @tie-rack Not to be too nitpicky on semantics, but shouldn't this be "referenda"?

tie-rack: @jungshadow some dictionaries put "referendums" first in the list of plurals for "referendum", so I'm feeling okay about it.

PR #34, opened 2015-03-25
Rename all the processing-migration files

tie-rack:
* The dates aren't meaningful or insightful
* Pick one naming style and apply it everywhere


7maples:
LGTM

PR #35, opened 2015-03-26
Validate no unreferenced rows

tie-rack:
Pivotal story: [90230424](https://www.pivotaltracker.com/story/show/90230424)

Validate that submitted CSVs do not contain records that are not referenced by any other files. Each record that has an ID that is not referenced by another file produces a warning.

(Also, the "full good run" test data has been replaced with a more recent VA special election data set, slightly modified to produce no warnings. The old data set, it turns out, was overflowing with unreferenced records.)


src/vip/data_processor/validation/db/reverse_references.clj
```diff
@@ -0,0 +1,61 @@
+(ns vip.data-processor.validation.db.reverse-references
+  (:require [vip.data-processor.validation.db.util :as util]
+            [korma.core :as korma]
+            [korma.sql.engine :as eng]))
+
+(defn find-referencing-column [table-name csv-spec]
+  (->> csv-spec
+       :columns
+       (filter #(= table-name (:references %)))
```
bwreid: I thought you didn't like this way of writing a function?  :)

src/vip/data_processor/validation/db/reverse_references.clj
```diff
@@ -0,0 +1,61 @@
...
+(defn join-clause [from-table from-column to-table]
+  (let [from-column (util/column-name from-table from-column)
+        to-column (util/column-name to-table "id")]
+    (list '= from-column to-column)))
+
+(defn where-clause-part [table column]
+  (let [column (util/column-name table column)]
+    (list '= column nil)))
+
+(defn find-unreferenced-rows [tables table-id references-map]
```
bwreid: I think this function should at least have a doc string -- although more is better IMO.

src/vip/data_processor/validation/db/reverse_references.clj
tie-rack: :sob: It's true.

src/vip/data_processor/validation/db/reverse_references.clj
```diff
@@ -0,0 +1,61 @@
...
+                 :let [table (:table csv-spec)
+                       references (find-referencing-tables table csv-specs)]
+                 :when (seq references)]
+             [table references])))
+
+(defn validate-no-unreferenced-rows-for-table [ctx [table-id references-map]]
+  (let [tables (:tables ctx)
+        unreferenced-rows (find-unreferenced-rows tables table-id references-map)]
+    (if (seq unreferenced-rows)
+      (assoc-in ctx [:warnings table-id "Unreferenced records"] unreferenced-rows)
```
bwreid: Is there a standard we've developed for these? My thought was they were all following: `[:higher-level-keyword "file-name" :error-type "additional details"]`

src/vip/data_processor/validation/db/reverse_references.clj
tie-rack: I'll add some!

src/vip/data_processor/validation/db/reverse_references.clj
tie-rack: You're right. We can find out from the `ctx` the file name the table was created for. I'll update that.

src/vip/data_processor/validation/db/reverse_references.clj
tie-rack: Is `(comp (partial = table-name) :references)` more or less readable? ;)

bwreid: "readable."

Looks okay to me.  :)

PR #36, opened 2015-03-30
[WR] validates that election, source, and state files include exactly one record

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/90231906)

Ensures that election.txt, source.txt, and state.txt only include one record. Otherwise, adds an error for each file.

Assigned @tie-rack.


tie-rack:
LGTM :shell:

PR #37, opened 2015-03-31
Validate no street segment overlaps

tie-rack:
Pivotal story: [90140090](https://www.pivotaltracker.com/story/show/90140090)

Validates that a CSV upload doesn't have any overlapping street segments. Adds errors for any such overlaps found.


bwreid:
Looks good but needs moar tests.

PR #39, opened 2015-04-02
Validate vip_id in source

tie-rack:
Pivotal story: [90232472](https://www.pivotaltracker.com/story/show/90232472)

Validate that the vip_id in source.txt is a valid FIPS code. If it's not, add an error.


bwreid:
Looks good to me!

PR #40, opened 2015-04-06
Check CSV values for non-UTF-8 characters

tie-rack:
Pivotal story: [90230268](https://www.pivotaltracker.com/story/show/90230268)

Add errors for any value with the Unicode replacement character (which means that there was an invalid UTF-8 byte sequence in the file there).


tank157:
LGTM

PR #41, opened 2015-04-07
[CS WR] validate complete addresses in election_administration.txt

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/90232754)

Checks for full physical address, full mailing address, or both. If either is incomplete, adds an error with the ID of the election_administration.

Assigned @7maples. Happy Birthday. :tada:


7maples:
lgtm

PR #43, opened 2015-04-10
Load XML elements with addresses

tie-rack:
Pivotal story: [91602732](https://www.pivotaltracker.com/story/show/91602732)

Load data from XML uploads into a SQLite database for the following elements:

* election_administration
* polling_location
* early_vote_site
* candidate
* street_segment


bwreid:
Looks :top: to me.

PR #48, opened 2015-04-27
[JE] output ballot line results to XML

7maples:
[91989656](https://www.pivotaltracker.com/story/show/91989656): Create `<ballot_line_result>` elements in XML export


tie-rack:
Awesome! Thank you!

PR #50, opened 2015-04-30
Validate generated XML

tie-rack:
Pivotal: [93444102](https://www.pivotaltracker.com/story/show/93444102)

After the QA Engine generates an XML export, it is validated against the VIP 3.0 spec. Failure to validate is added as a fatal error.


bwreid:
:eyes: :+1: :two: :man:

PR #51, opened 2015-05-05
[WR] upload xml to an s3 bucket

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/91968202)

Uploads the xml generated at the end of the run to a configured S3 bucket so that it can later be accessed.

### DIFF
--- a/src/vip/data_processor/s3.clj
+++ b/src/vip/data_processor/s3.clj
@@ -50,4 +50,4 @@
   (let [new-filename (xml-filename ctx)]
     (put-object (str "processed-feeds/" new-filename)
                 (.toFile (:xml-output-file ctx)))
-    ctx))
+    (assoc ctx :generated-xml-filename new-filename)))
--- a/src/vip/data_processor/s3.clj
+++ b/src/vip/data_processor/s3.clj
@@ -1,7 +1,9 @@
 (ns vip.data-processor.s3
   (:require [aws.sdk.s3 :as s3]
             [clojure.java.io :as io]
-            [turbovote.resource-config :refer [config]])
+            [turbovote.resource-config :refer [config]]
+            [korma.core :as korma]
+            [clojure.string :refer [join]])
   (:import [java.nio.file Files CopyOption StandardCopyOption]
            [java.nio.file.attribute FileAttribute]))
 
@@ -9,6 +11,11 @@
   (s3/get-object (config :aws :creds)
                  (config :aws :s3 :bucket)
                  key))
+
+(defn put-object [key value]
+  (s3/put-object (config :aws :creds)
+                 (config :aws :s3 :bucket)
+                 key value))
 
 (def tmp-path-prefix "vip-data-processor")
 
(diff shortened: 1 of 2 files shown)


src/vip/data_processor/s3.clj
```diff
@@ -23,3 +30,28 @@
                 tmp-path
                 (into-array [StandardCopyOption/REPLACE_EXISTING]))
     tmp-path))
+
+(defn format-fips-code [fips]
```
tank157: I think this can all be accomplished simpler with format:

https://clojuredocs.org/clojure.core/format

Something like (format "%05d" fips) should do the trick.

src/vip/data_processor/s3.clj
```diff
@@ -23,3 +30,28 @@
...
+            padding (apply str (repeat padding-length "0"))]
+        (str padding fips))
+      fips)))
+
+(defn xml-filename [ctx]
+  (let [fips (-> (get-in ctx [:tables :sources])
+                 korma/select first :vip_id format-fips-code)
+        election-date (->> (get-in ctx [:tables :elections])
+                           korma/select first :date)
+        timestamp (.getTime (java.util.Date.))]
```
tank157: The timestamp will be in Unix time, is that what is intended? It will cut through TimeZone issues, but isn't the most readable format.

src/vip/data_processor/s3.clj
```diff
@@ -23,3 +30,28 @@
...
+    (->> (vector "vipFeed" fips election-date timestamp)
```
tank157: Seems you could also just do `(join "-" ["vipFeed" fips election-date timestamp])`

src/vip/data_processor/s3.clj
```diff
@@ -23,3 +30,28 @@
...
+         (join "-"))))
+
+(defn upload-to-s3
+  "Uploads the generated xml file to the specified S3 bucket."
+  [ctx]
+  (let [new-filename (xml-filename ctx)]
+    (put-object (str "processed-feeds/" new-filename)
+                (slurp (str (:xml-output-file ctx))))
```
tank157: So while str is doing the right thing here, giving us the fully qualified path name to the file, it seems hacky. I wonder if there's actually a real File method we could call instead to make it clearer what is going on here.

I think the .getCanonicalPath java function would work here and is clearer. Maybe give that a try?

src/vip/data_processor/s3.clj
bwreid: I had tried something like this but it seems like it only accepts digits. The only comparable code I found was this but more complex. If you're able to reproduce something different though, please do let me know.  :)

tank157: You could do
`(format "%05d" (read-string "12"))  ;=> "00012"`

tank157: Though if some non-integer fips code sneak in somewhere up the line and aren't caught, then this will break.

src/vip/data_processor/s3.clj
bwreid: Yup, Noma confirmed this works.

src/vip/data_processor/s3.clj
tie-rack: Sorry to poke my head in here, but `slurp` is going to read the entire file into memory, and these files may be too large to fit in memory. Fortunately, `put-object` can also take an InputStream or File, so the reference to the file itself (`(:xml-output-file ctx)`) may just work!

Back to :palm_tree: :ocean:

tank157:
LGTM

PR #52, opened 2015-05-07
Init processing table

bwreid:
[Pivotal Card 1](https://www.pivotaltracker.com/story/show/91446926)
[Pivotal Card 2](https://www.pivotaltracker.com/story/show/91447160)

This work:

1. Initializes a processing table called `results` 
1. Loads basic information about the run to that table such as start and end time.

Also added code to `dev.core` and the `project.clj` that allows for a `dev-resources/config.edn` file to be included so this can be run locally. The config file only needs to be super simple and should look like:
```
{:postgres {:host "127.0.0.1"
            :port "5432"
            :user "YOUR PSQL USERNAME"
            :password "YOUR PSQL PASSWORD"}}
```
If you're not sure of this information but have access to the `psql` command, run it and then run `\conninfo` and you'll see how you're logged in.

### DIFF
--- a/dev-src/dev/core.clj
+++ b/dev-src/dev/core.clj
@@ -16,14 +16,14 @@
            t/attach-sqlite-db
            (data-spec/add-data-specs data-spec/data-specs)
            t/xml-csv-branch]
+          [psql/start-run]
           db/validations
           xml-output/pipeline))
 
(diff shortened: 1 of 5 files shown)


ericnormand:
LGTM

PR #53, opened 2015-05-08
Init validations table

bwreid:
[Pivotal Card 1](https://www.pivotaltracker.com/story/show/91446950)
[Pivotal Card 2](https://www.pivotaltracker.com/story/show/91447238)

This PR allows for uploading of validation errors to the postgresql database. All remaining content after identifying the severity, source, and description is placed in message.

### DIFF
--- a/src/vip/data_processor/db/postgres.clj
+++ b/src/vip/data_processor/db/postgres.clj
@@ -61,13 +61,13 @@
                                  :description (if (keyword? description)
                                                 (name description)
                                                 description)
-                                 :message (apply str message)}))))
+                                 :message (pr-str message)}))))
 
 (defn insert-validations [{:keys [warnings errors critical fatal] :as ctx}]
   (let [result-id (:import-id ctx)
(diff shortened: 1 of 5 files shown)


src/vip/data_processor/db/postgres.clj
```diff
@@ -42,3 +47,29 @@
...
+    (korma/insert validations
+                  (korma/values {:result_id id
+                                 :severity (name severity)
+                                 :scope (if (keyword? scope-key)
+                                          (name scope-key)
+                                          scope-key)
+                                 :description (if (keyword? description)
+                                                (name description)
+                                                description)
+                                 :message (apply str message)}))))
```
tie-rack: Should this be `:message (pr-str message)`? So that it goes in as edn?

Especially if these things can be vectors. `(apply str [1 2 3])` is `"123"`, while `(pr-str [1 2 3])` is `"[1 2 3]"`.

tie-rack:
Okay. This LGTM!

PR #54, opened 2015-05-08
Refine validation reporting

bwreid:
There is no Pivotal Card for this PR -- however, this is an attempt to get all the validation errors we report into a similar format so that we can expect something more specific when we reach the message column.


tie-rack:
Yeah. Good call. Nice work.

tie-rack:
Smart work. I like it. LGTM.

PR #55, opened 2015-05-14
Speed up XML generation

tie-rack:
Optimize some slowness.

Also adds some logging to give some feedback while running it via `lein run`.

### DIFF
--- a/src/vip/data_processor/output/xml.clj
+++ b/src/vip/data_processor/output/xml.clj
@@ -120,13 +120,12 @@
           counter (atom 0)]
       (.write out-file "<?xml version='1.0' encoding='UTF-8'?>\n")
       (.write out-file opening-tag)
-      (doseq [xml-node-fn xml-node-fns]
-        (let [xml-children (xml-node-fn ctx)]
-          (doseq [xml-child xml-children]
-            (swap! counter inc)
-            (when (zero? (mod @counter 1000))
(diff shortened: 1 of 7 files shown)


src/vip/data_processor/output/xml.clj
```diff
@@ -60,47 +82,54 @@
...
-                          with-out-str
-                          (string/replace #"(>)\n|\n(<)" "$1$2"))]
-          (.write out-file xml-str)))
+      (doseq [xml-node-fn xml-node-fns]
+        (let [xml-children (xml-node-fn ctx)]
+          (doseq [xml-child xml-children]
+            (swap! counter inc)
+            (when (zero? (mod @counter 1000))
+              (log/info "Wrote" @counter "XML nodes"))
+            (emit-element out-file xml-child))))
```
ericnormand: let-binding the xml-children will hold onto the head unnecessarily. If you get rid of it, you also can collapse the two `doseq`s.


```
(doseq [xml-node-fn xml-node-fns
             xml-child xml-children]
          (swap! counter inc)
          (when (zero? (mod @counter 1000))
              (log/info "Wrote" @counter "XML nodes"))
          (emit-element out-file xml-child))
```

tie-rack: You're right! I'll get that in tomorrow.

bwreid:
Post Eric's suggested change, lgtm.

PR #56, opened 2015-05-27
[CS WR] zip up the final xml

bwreid:
[Pivotal Card 94345396](https://www.pivotaltracker.com/story/show/94345396)

This PR allows us to zip the final XML that is outputted by data-processor.

### DIFF
--- a/src/vip/data_processor/s3.clj
+++ b/src/vip/data_processor/s3.clj
@@ -4,8 +4,12 @@
             [turbovote.resource-config :refer [config]]
             [korma.core :as korma]
             [clojure.string :refer [join]])
-  (:import [java.nio.file Files CopyOption StandardCopyOption]
-           [java.nio.file.attribute FileAttribute]))
+  (:import [java.io File]
+           [java.nio.file Files CopyOption StandardCopyOption]
+           [java.nio.file.attribute FileAttribute]
+           [net.lingala.zip4j.core ZipFile]
+           [net.lingala.zip4j.model ZipParameters]
+           [net.lingala.zip4j.util Zip4jConstants]))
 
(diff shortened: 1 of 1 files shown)


src/vip/data_processor/s3.clj
```diff
@@ -47,7 +51,13 @@
...
-    (put-object (str "processed-feeds/" new-filename)
-                (.toFile (:xml-output-file ctx)))
-    (assoc ctx :generated-xml-filename new-filename)))
+  (let [zip-filename (str (xml-filename ctx) ".zip")
+        xml-file (-> ctx :xml-output-file .toFile)
+        zip-file (ZipFile. zip-filename)
+        zip-params (doto (ZipParameters.)
+                     (.setCompressionLevel
+                      Zip4jConstants/DEFLATE_LEVEL_NORMAL))]
+    (.createZipFile zip-file xml-file zip-params)
```
tank157: So is this going to write a file, and if so, do we need to do any cleanup? I assume it's in a tmp folder type situation, but even those can fill up.

bwreid: If so, wouldn't we need to do the same for the .xml file? If that's the case, we may want to add a step to clean up all generated files.

tank157: Yeah, I was sort of thinking for both files, I should have made that clear.

tank157:
LGTM with the file cleanup taken care of in another PR.

src/vip/data_processor/s3.clj
bwreid: Created [this Pivotal Card](https://www.pivotaltracker.com/story/show/95561156) to address removing these files.

PR #57, opened 2015-06-03
Load zip or xml

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/94878640)

This PR allows for us to use direct XMLs or Zip files to run the data-processor. It will be especially helpful when running locally.

### DIFF
--- a/dev-src/dev/core.clj
+++ b/dev-src/dev/core.clj
@@ -12,7 +12,7 @@
             [democracyworks.squishy.data-readers]))
 
 (def pipeline
-  (concat [zip/unzip
+  (concat [zip/assoc-file
            zip/extracted-contents
            t/attach-sqlite-db
            (data-spec/add-data-specs data-spec/data-specs)
@@ -24,7 +24,9 @@
 
 (defn -main [zip-filename]
   (psql/initialize)
-  (let [zip (java.io.File. zip-filename)
+  (let [zip (if (zip/xml-file? zip-filename)
+              (java.nio.file.Paths/get zip-filename (into-array [""]))
+              (java.io.File. zip-filename))
         result (pipeline/process pipeline zip)]
     (when-let [xml-output-file (:xml-output-file result)]
       (println "XML:" (.toString xml-output-file))
--- a/dev-src/dev/core.clj
+++ b/dev-src/dev/core.clj
@@ -22,12 +22,12 @@
           xml-output/pipeline
           [psql/insert-validations]))
 
-(defn -main [zip-filename]
+(defn -main [filename]
   (psql/initialize)
-  (let [zip (if (zip/xml-file? zip-filename)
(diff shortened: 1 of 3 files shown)


src/vip/data_processor/validation/zip.clj
```diff
@@ -15,17 +20,22 @@
...
 
-(defn unzip [ctx]
+(defn assoc-file [ctx]
   (let [path (:input ctx)]
     (if (zip-file? path)
       (assoc ctx :input (unzip-file path))
-      (assoc ctx :stop (str path " is not a zip file!")))))
+      (if (xml-file? path)
+        (assoc ctx :input path)
+        (assoc ctx :stop (str path " is not a zip file!"))))))
```
tie-rack: I think maybe a `cond` here would be more readable than nested `if`s. Agree?

```clojure
(cond
  (zip-file? path) (assoc ctx :input (unzip-file path))
  (xml-file? path) (assoc ctx :input path)
  :else (assoc ctx :stop (str path " is not a zip or xml file!")))
```

Also, that stop message should change since unzipped XML files are now acceptable.

src/vip/data_processor/validation/zip.clj
```diff
@@ -15,17 +20,22 @@
...
   (let [path (:input ctx)]
-    (assoc ctx :input
-           (-> path
-               (.resolve "data")
-               .toFile
-               .listFiles
-               seq))))
+    (if (xml-file? path)
+      (assoc ctx :input
+             (seq [(.toFile path)]))
```
tie-rack: Is it necessary to call `seq` on the vector? It's required down below because `.listFiles` returns a java array and we need something that implements `ISeq` (for `first` and `map`), but vectors already do.

dev-src/dev/core.clj
```diff
@@ -24,7 +24,9 @@
 
 (defn -main [zip-filename]
   (psql/initialize)
-  (let [zip (java.io.File. zip-filename)
+  (let [zip (if (zip/xml-file? zip-filename)
+              (java.nio.file.Paths/get zip-filename (into-array [""]))
+              (java.io.File. zip-filename))
```
tie-rack: I'm afraid I'm confused (and it might be my fault). Why does `zip` (which could now be either a zip file or an xml file) need to be a `Path` object if it's xml and a `File` object if it's a zip?

src/vip/data_processor/validation/zip.clj
bwreid: Good call. Thank you.

dev-src/dev/core.clj
bwreid: I'll correct the naming. The reason is because if it's an XML we're calling .toFile on it later. And it's not happy when you call .toFile on a File.  :(

tie-rack: Update this name to something and we've got a LGTM.

PR #58, opened 2015-06-09
[CS] Import from sqlite to postgres at the end

tie-rack:
* Imports every table except street segments into Postgres
* Differences from sqlite tables: `result_id` column referencing the `id` from the `results` table on every table, each table's primary keys is `(results_id, id)`, uses Postgres's boolean type (where applicable).
* Coerces some data from how they come out of SQLite for inserting into Postgres (integers, dates, and booleans)
* Move and refactor some shared SQL code

After running an import, connect to Postgres, find the `id` from the `results` table, and you'll be able to run queries like (e.g.) `SELECT * FROM candidates WHERE results_id = 1`.

Pivotal story: [95050110](https://www.pivotaltracker.com/story/show/95050110)


bwreid:
Overall, I think this looks great. Thanks, Chris!

PR #61, opened 2015-06-16
Track temp files and delete them in the end

tie-rack:
Pivotal story: [95561156](https://www.pivotaltracker.com/story/show/95561156)

Adds a function to the end of the pipeline that cleans up temporary files created during a run (download, extracted files, generated XML), so as to take some pressure off the OS from doing it itself.


bwreid:
LGTM

PR #62, opened 2015-06-17
Create "public_id" for each run

tie-rack:
Pivotal story: [96666240](https://www.pivotaltracker.com/story/show/96666240)

Creates a new column inside of the results table that contains a more readable _unique_ ID that's generated based on the election date, the election type, and the state. The unique identifier ends with the row's id itself.

### DIFF
--- a/src/vip/data_processor/db/postgres.clj
+++ b/src/vip/data_processor/db/postgres.clj
@@ -47,6 +47,13 @@
                                              :complete false}))]
     (assoc ctx :import-id (:id results))))
 
+(defn build-public-id [date election-type state import-id]
+  (let [nil-or-empty? (some-fn nil? empty?)
+        good-parts (remove nil-or-empty? [date election-type state])]
+    (if (empty? good-parts)
+      (str "invalid-" import-id)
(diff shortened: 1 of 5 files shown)


src/vip/data_processor/db/postgres.clj
```diff
@@ -46,6 +47,26 @@
...
+  (let [state (-> ctx
+                 (get-in [:tables :states])
+                 (korma/select (korma/fields :name))
+                 first
+                 :name)
+        {:keys [date election_type]} (-> ctx
+                                         (get-in [:tables :elections])
+                                         (korma/select (korma/fields :date :election_type))
+                                         first)]
+    (str/join "-" [date election_type state import-id])))
```
bwreid: If one or more of these things are missing, what will happen?

tie-rack: Good call. It would end up looking like (if the election-type were missing): "2014-04-10--Ohio-10", with two adjacent hyphens. I can collapse the hyphens around missing parts (Fortunately, the final piece _cannot_ be missing, so there will _always_ be a unique public_id.)

bwreid:
LGTM!

PR #63, opened 2015-06-19
Allow much larger ids in postgres db

tie-rack:
Some VIP users generate ids for their data that are larger than Postgres's INTEGER type. This changes all id fields (and their references) to be BIGINTs to accommodate that.


resources/migrations/20150605-26-custom-ballot-ballot-responses.up.sql
```diff
@@ -1,4 +1,4 @@
-CREATE TABLE custom_ballot_ballot_responses (results_id INTEGER REFERENCES results (id) NOT NULL,
-                                             custom_ballot_id INTEGER,
-                                             ballot_response_id INTEGER,
+CREATE TABLE custom_ballot_ballot_responses (results_id BIGINT REFERENCES results (id) NOT NULL,
+                                             custom_ballot_id BIGINT,
+                                             ballot_response_id BIGINT,
                                              sort_order INTEGER);
```
bwreid: If we're changing _everything_ to BIGINT should we do this as well? Otherwise, we likely don't need to change things like 'total_votes' to BIGINT.

tie-rack: Oops. Yeah, I changed everything and then saw `sort_order` and thought, we don't need _that_ to be gigantic. This is really about ids, so I'll undo the non-id columns.

tie-rack: Okay, the non-id columns that got BIGINTified have been re-INTEGERed.

bwreid:
Sweet. LGTM.

PR #64, opened 2015-06-19
Friday itch scratching

tie-rack:
* Reduces the noise while running the processor (especially helpful for huge imports with tons of errors)
* Updates the dev processor pipeline to add a row to the results table right away instead of waiting until after all the data has been loaded


bwreid:
I like it. LGTM.

PR #65, opened 2015-06-25
Reformat validation errors

tie-rack:
1. The validations table structure has been updated to: results_id, severity, scope, identifier, error_type, error_data
2. All errors follow the format detailed above
3. The way the statistics table is populated has been updated based on the new error format
4. Tests are updated to pass correctly with these changes implemented

Pivotal story: [97112626](https://www.pivotaltracker.com/story/show/97112626)


src/vip/data_processor/db/postgres.clj
```diff
@@ -87,38 +88,46 @@
...
-    (when errors (insert-severity-fn 'errors errors))
-    (when critical (insert-severity-fn 'critical critical))
-    (when fatal (insert-severity-fn 'fatal fatal))
-    ctx))
+(def global-identifier -1)
+
+(defn identifier->int [identifier]
+  (cond
+    (= :global identifier) global-identifier
+    (string? identifier) (BigDecimal. identifier)
```
tank157: BigDecimal is an int now?
Anything that's not global nor a string must be an int?

tie-rack: Good point. This fn changed without the name getting updated. I'll update the name.

As far as anything not `:global` or a string: if it's something not able to be coerced into a Postgres bigint or `nil`, then it's a problem for the caller, enforced later by Postgres itself. Of course, with this code base, it's safe... for now... though perhaps a docstring and precondition would be helpful.

src/vip/data_processor/db/postgres.clj
```diff
@@ -87,38 +88,46 @@
...
+(defn validation-value [results-id severity scope identifier error-type error-data]
+  {:results_id results-id
+   :severity (name severity)
+   :scope (name scope)
+   :identifier (identifier->int identifier)
+   :error_type (name error-type)
+   :error_data (pr-str error-data)})
+
+(defn validation-values [{:keys [import-id] :as ctx}]
+  (mapcat
```
tank157: A docstring with a sample of the format of the errors data before/after flatten keys would be nice here, because that inner mapcat destructuring is a bit confusing.

src/vip/data_processor/db/statistics.clj
```diff
@@ -13,16 +14,13 @@
...
-          (cond
-            (map? error-or-errors) (+ acc (reduce count-errors 0 (vals error-or-errors)))
-            (coll? error-or-errors) (+ acc (count error-or-errors))
-            (nil? error-or-errors) acc
-            :else (inc acc)))]
-    (->> [warnings errors critical fatal]
-         (map table-name)
-         (map (partial count-errors 0))
-         (reduce + 0))))
+  (transduce (comp (map table-name)
```
tank157: Nice use of transducers. I have to say, though, coming from the outside, this is pretty opaque. Is `table-name` a good identifier here? It seems like it might actually be where the data is coming from, but it's hard to tell without knowing the structure of the data-spec offhand.

tie-rack: It really is the table's name (as a keyword). For example, `:elections`.

tank157: Ah, ok, I had it in my head that the warnings, errors, etc were also keywords for some reason, but that must be the data.

tank157:
LGTM

PR #67, opened 2015-07-13
Be more memory conscious

tie-rack:
* `db.util/select-*-lazily` loads rows from a database as a chunked lazy sequence
* Importing from sqlite to postgres now uses `select-*-lazily` so that it doesn't pull gigantic tables into memory
* Validating CSV value formats and #-of-values violations now happens as rows from the CSV are being read. Before, we had been inadvertently doing those checks by realizing the entirety of the CSV in memory, which was causing problems for very large files.
* Limits for how many XML elements to read in at a time was raised, as it was too small.

Fixes bug [98585266](https://www.pivotaltracker.com/story/show/98585266)

### DIFF
--- a/src/vip/data_processor/db/postgres.clj
+++ b/src/vip/data_processor/db/postgres.clj
@@ -134,7 +134,7 @@
                    errors)))))
    [:warnings :errors :critical :fatal]))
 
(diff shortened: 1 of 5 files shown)


bwreid:
Great. :+1:

PR #68, opened 2015-07-13
Handle nil vip_id when creating xml filename

tie-rack:
It was blowing up when trying to generate a filename for feeds missing a `vip_id`.

Fixes bug [98939106](https://www.pivotaltracker.com/story/show/98939106)


bwreid:
Capital L.G.T.M.

PR #69, opened 2015-07-15
Remove NOT NULL constraints on non-id columns

tie-rack:
There were NOT NULL constraints on columns that are either NULLable, or are often provided in imports without values.

The "processing-migrations" migrations don't need a separate migration to remove the constraint and can be edited in place because they are run on a fresh database for each import.

Pivotal bug [99134060](https://www.pivotaltracker.com/story/show/99134060)


bwreid:
LGTM.

PR #70, opened 2015-07-15
Allow zip files to not include "data" dir

tie-rack:
Broaden the structure of zip files we handle to more accurately follow what we're getting.

Pivotal bug [98969682](https://www.pivotaltracker.com/story/show/98969682)


bwreid:
Looks good.  :)

PR #71, opened 2015-07-22
Fix import to Postgres when value types are bad

tie-rack:
If a feed has a value that isn't an integer in a field that requires an integer, the import would fail when importing the data into postgres. Non-integers (and non-dates for date fields, and non-yes/no for boolean fields) will no longer throw an exception and instead insert nil into the database. An error is already logged for the type mismatch.

Pivotal bug [99541752](https://www.pivotaltracker.com/story/show/99541752)


bwreid:
Nice. This looks great to me.

PR #72, opened 2015-07-22
Make value checks case insensitive

tie-rack:
Lower-case values before checking against a list of lower-cased options.

Pivotal bug [99533180](https://www.pivotaltracker.com/story/show/99533180)


bwreid:
LGTM.

PR #73, opened 2015-07-22
Better validate electoral district types

tie-rack:
Use the "official" electoral district type regex.

Pivotal story: [99558306](https://www.pivotaltracker.com/story/show/99558306)


bwreid:
Sweet tests. Looks good to me.

PR #74, opened 2015-07-23
[WR] add exception column to results table and fill on fail

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/98991978)

What do you think this as a start? We'd add the cause of the exception to the results table for it to then be shown on the Metis side.

### DIFF
--- a/resources/migrations/20150723-00-add-exception-to-results.down.sql
+++ b/resources/migrations/20150723-00-add-exception-to-results.down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE results DROP COLUMN exception text;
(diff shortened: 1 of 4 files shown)


tie-rack:
I'm not sure about actually showing the exception to the user, but this is helpful. What about adding an end time to the results table?

bwreid:
I was initially starting with that but the card says we should add as much detailed information as possible -- which makes sense. Just saying the feed failed doesn't really help. Unless we can assume we'll always be handling anything that throws an exception, in which case it may not matter. I'll check with Maria.  :)

tie-rack:
LGTM

PR #75, opened 2015-07-27
Lower sqlite statement parameter limit to 500

tie-rack:
500 is the SQLITE_MAX_COMPOUND_SELECT limit, which is being surpassed somehow in some inserts. Perhaps there's a way to know when to use the SQLITE_MAX_COMPOUND_SELECT and when to use the SQLITE_MAX_VARIABLE_NUMBER (999), but until then... this.


tank157:
LNCTM (looks non-controversial to me).

PR #76, opened 2015-07-28
Add public-id to RabbitMQ complete message

tie-rack:
Needed by notifications work (see: Pivotal story [98952286](https://www.pivotaltracker.com/story/show/98952286)).


bwreid:
LGTM.

PR #77, opened 2015-08-04
[WR] removes type validations from election and localities

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/99557896)

Removes type checking for Elections and Localities.


tie-rack:
Easy. LGTM.

PR #78, opened 2015-08-20
[WR] remove timestampe from the zip filename

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/101053518) Removes the timestamp from the generated zips filename.


tie-rack:
Mmmmmmyup. LGTM.

PR #79, opened 2015-08-24
[WR]  feeds come and go to different aws buckets

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/101053422)

Separates the buckets so that unprocessed feeds will now be separate from processed ones that we create.


tie-rack:
Looks great!

tie-rack:
Yes. Nice.

PR #80, opened 2015-08-31
[WR] remove validation from contest type

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/101897132)

Updates data-spec to not worry about formatting for contest type.


tie-rack:
LGTM

PR #81, opened 2015-08-31
[CS] Clean up zip file after it's uploaded to S3

tie-rack:
Fixes both a disk space leak and a bug where a processor wouldn't be able to process a second feed for the same election.


bwreid:
LGTM

PR #82, opened 2015-09-03
Check for in-file dupe ids during import

tie-rack:
Duplicate ids _within_ a file were causing the CSV importer to blow up as it tried to insert values into a table. Now, we try the insert, catch an error, and if it's because of a UNIQUE constraint on an id column, it retries, but without any ids already in the table or within the chunk it's working on.

This results in rows not being imported, so it adds a fatal error for each duplicated id.

[Pivotal story](https://www.pivotaltracker.com/story/show/101913786)


bwreid:
Looks Great To Me.

PR #83, opened 2015-09-03
Safe sqlite db path

tie-rack:
Use the import's id as the basis for the SQLite db path instead of the original filename so we don't run into a java.net.URISyntaxException.

[Pivotal story](https://www.pivotaltracker.com/story/show/101067048)


bwreid:
LGTM.

PR #84, opened 2015-09-04
Handle XMLStreamException during xml import

tie-rack:
Move to explicit looping so we can catch a XMLStreamException while maintiaining any data we've imported up to that point.

[Pivotal story](https://www.pivotaltracker.com/story/show/99941006)


bwreid:
This is a really clever solution. LGTM.

PR #85, opened 2015-09-08
remove invalid extensions

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/99943928)

Checks for any files within the feed that is not a csv, txt, or xml and removes those files from the :input.


tie-rack:
Should this function also add a warning about whatever files it removes?

bwreid:
Survey (Maria) says: Yes. I'll do that next.

PR #87, opened 2015-09-10
Duplicate ID checking for XML

tie-rack:
Required moving a helper function and, more importantly, changing bulk-import to take and return a context for retrying inserts with duplicate ids and to be able to add errors when there are duplicate ids.


bwreid:
Looks Great To Me.

PR #88, opened 2015-09-17
Add New Relic

tie-rack:
* Add newrelic.yml configuration file
* Add com.newrelic.agent.java/newrelic-agent java agent
* Copy java agent on build with jar-copier (so it's available in the Docker image from the version downloaded from `lein deps`)
* Update Dockerfile CMD to use the java agent and set the newrelic.environment property (switches to the shell version of CMD, so that system property can be set from an environment variable)
* Update @.service.template file to pull necessary environment variables from Consul
* Add clj-newrelic to dependencies
* Use `defn-traced` for various functions we might be interested in tracing

Pivotal story: [102003328](https://www.pivotaltracker.com/story/show/102003328)


bwreid:
This looks great.

PR #89, opened 2015-09-22
[WR] street segments require necessary address elements

bwreid:
A first step to fixing some validation problems with street segments.


tie-rack:
LGTM :)

PR #90, opened 2015-09-23
[CS] Send logs to Loggly (if available)

tie-rack:
* Move to Log4j2 (for proprty substitution in configuration)
* Configure Log4j2 with a syslog appender
* Link a "loggly" container as "syslog" in @.service.template (this will set the required environment variables used in the configuration)

Pivotal story: [102023108](https://www.pivotaltracker.com/story/show/102023108)

### DIFF
--- a/data-processor@.service.template
+++ b/data-processor@.service.template
@@ -6,6 +6,8 @@ After=consul@%i.service
 Wants=consul@%i.service
 After=rabbitmq@%i.service
 Requires=rabbitmq@%i.service
+After=loggly@%i.service
+Requires=loggly@%i.service
 
 [Service]
 EnvironmentFile=/etc/environment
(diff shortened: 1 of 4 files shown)


cap10morgan:
Just curious: Why not logback? I was under the impression that "just use logback" was the advice du jour for JVM logging.

tie-rack:
@cap10morgan We were already using Log4j. And the Log4j2 about page throws some shade at Logback: "Log4j 2... provides many of the improvements available in Logback while fixing some inherent problems in Logback's architecture"

In either case, they are both widely used and this was an upgrade, not a lateral move.

cap10morgan:
Gotcha, makes sense. The logging library wars will rage on it seems.

bwreid:
Shiny! So clean. LGTM.

PR #92, opened 2015-10-02
Don't try to insert nothing

tie-rack:
Fixes a bug where if, like, an entire bulk import chunk is nothing but duplicate ids, it doesn't try to insert nothing.


bwreid:
:+1: Looks Good To Me

PR #93, opened 2015-10-06
Strip non-word characters from header names

tie-rack:
This ends up fixing a problem where a byte order marker would be read as part of a header's name, and should also help other classes of header mistakes (extraneous spaces, accidental hyphens).

Pivotal story: [103447542](https://www.pivotaltracker.com/story/show/103447542)


bwreid:
This is great. LGTM.

PR #94, opened 2015-10-06
[WR] add incorrectly formatted value to errors

bwreid:
[Pivotal Card](https://www.pivotaltracker.com/story/show/101985558)

The incorrect value will make its way to the front-end as well!


tie-rack:
Amazing. LGTM.

PR #95, opened 2015-10-06
Add exception message to processing.complete

tie-rack:
Moves when the exception that stops processing is thrown, so that it happens after the "processing.complete" message would be sent to rabbit.

An exception needs to be thrown so that the SQS message that kicked off processing can be put on the fail queue.


bwreid:
Cool use of `cond->`. LGTM!

PR #96, opened 2015-10-06
Skip a BOM when reading files (XML and CSV)

tie-rack:


src/vip/data_processor/util.clj
```diff
@@ -21,3 +23,17 @@
...
+
+(defn bom-safe-reader
+  "Returns a reader from io/reader, which has advanced past a byte
+  order marker if one exists."
+  [x & opts]
+  (let [reader (apply io/reader x opts)]
+    (.mark reader 10)
+    (let [first-char (.read reader)]
+      (if (= first-char BOM)
+        (log/info x "had byte order marker")
```
bwreid: Is it worth having this be a warning / error / something? Partners will likely not be able to fix this but it may be nice to report it as a non-blocking problem.

src/vip/data_processor/validation/csv.clj
```diff
@@ -1,12 +1,12 @@
 (ns vip.data-processor.validation.csv
   (:require [clojure.data.csv :as csv]
-            [clojure.java.io :as io]
             [clojure.set :as set]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
             [com.climate.newrelic.trace :refer [defn-traced]]
             [korma.core :as korma]
             [korma.db :as db]
+            [vip.data-processor.util :as bom]
```
bwreid: Can we do `[vip.data-processor.util :as util]` like we do below?

src/vip/data_processor/util.clj
tie-rack: I'm not sure. It's not actually a problem on their end, or a problem at all, really. A BOM is a nice thing that we're somehow reading wrong.

src/vip/data_processor/validation/csv.clj
tie-rack: `vip.data-processor.db.util` is aliased as `util`. But I'll change that alias to `db.util` and this one to `util`. That's better.

bwreid: Ah great, thanks.

src/vip/data_processor/util.clj
bwreid: Alright, good enough for me.  :)

PR #98, opened 2015-10-09
Make CSV filename matching case insensitive

tie-rack:
Allow CSVs to be named "election.txt", "Election.txt", "ELECTION.TXT", or whatever.

Pivotal story: [103614990](https://www.pivotaltracker.com/story/show/103614990)


bwreid:
Nice and easy. Well done.

PR #99, opened 2015-10-13
[CS] vip_id is a string of digits, not a number

tie-rack:
Fixes a problem where low a vip_id would be converted to a number ("01" -> 1), causing a validation to fail (bad vip_id) erroneously.

Pivotal bug: [105597372](https://www.pivotaltracker.com/story/show/105597372)


bwreid:
Looks good. Merge!

PR #100, opened 2015-10-15
Use id in format rule check if available

tie-rack:
If an element being checked with a format rule has an `"id"` key, use that as the identifier, otherwise use whatever was passed as `id-or-line-number` (which will presumably be the line number at that point).

Pivotal bug: [105713556](https://www.pivotaltracker.com/story/show/105713556)


bwreid:
Looks Great To Me.

PR #101, opened 2015-10-26
Election id approval

ericnormand:
We need to store a new piece of information called the "election-id". It uniquely identifies an election across all imports. This election-id can be tagged "approved" which we store in the database in a table called election_approvals. We reference that table from the results table by election-id.

[Pivotal Card](https://www.pivotaltracker.com/story/show/103355828)

### DIFF
--- a/src/vip/data_processor/db/postgres.clj
+++ b/src/vip/data_processor/db/postgres.clj
@@ -58,17 +58,35 @@
       (str "invalid-" import-id)
       (str/join "-" (concat good-parts [import-id])))))
 
-(defn generate-public-id [{:keys [import-id] :as ctx}]
+(defn build-election-id [date election-type state]
+  (let [components [date election-type state]]
+    (when (every? seq components)
+      (->> components
+           (map str/trim)
+           (str/join "-")))))
+
+(defn get-public-id-data [{:keys [import-id] :as ctx}]
   (let [state (-> ctx
-                 (get-in [:tables :states])
-                 (korma/select (korma/fields :name))
-                 first
-                 :name)
+                  (get-in [:tables :states])
+                  (korma/select (korma/fields :name))
+                  first
+                  :name)
         {:keys [date election_type]} (-> ctx
                                          (get-in [:tables :elections])
                                          (korma/select (korma/fields :date :election_type))
                                          first)]
-    (build-public-id date election_type state import-id)))
+    {:date date
+     :election-type election_type
+     :state state
+     :import-id import-id}))
+
+(defn generate-public-id [ctx]
+  (let [{:keys [date election-type state import-id]} (get-public-id-data ctx)]
+    (build-public-id date election-type state import-id)))
+
+(defn generate-election-id [ctx]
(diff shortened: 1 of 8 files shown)


src/vip/data_processor/db/postgres.clj
```diff
@@ -58,17 +59,35 @@
       (str "invalid-" import-id)
       (str/join "-" (concat good-parts [import-id])))))
 
-(defn generate-public-id [{:keys [import-id] :as ctx}]
+(defn build-election-id [date election-type state]
+  (let [components [date election-type state]]
+    (when (every? seq components)
+      (->> components
+           (map str/trim)
```
tank157: I don't know how picky we need to be here, because I don't really know the rest of this codebase, but should the trimming happen before the `(every? seq ...)` call? That way, if a component had a value of " " it would trigger the "not there" condition, rather than ending up with an id with "--"?

It looks like there's a similar problem in the `build-public-id`, assuming this is a potential problem. Perhaps there are checks in other places in the code that would prevent this?

src/vip/data_processor/db/postgres.clj
```diff
@@ -78,6 +97,24 @@
...
+     (when-not (seq (korma/select election_approvals
+                                  (korma/where {:election_id election-id})))
+       (korma/insert election_approvals
+                     (korma/values {:election_id election-id}))))))
+
+(defn store-election-id [ctx]
+  (if-let [election-id (generate-election-id ctx)]
+    (let [id (:import-id ctx)]
+      (save-election-id! election-id)
+      (korma/update results
```
tank157: I suppose there is the possibility of inserting into the `election_approvals` table and not updating the `election_id` if the server goes down here, because there isn't a transaction around them both (unless there's a higher one?) But it seems  unlikely, so I'm not overly concerned about it, and I'd presume we'd re-run any processes that it died partway, so the next one would just skip `save-election-id!` bit.

Is that roughly the compromise you are making here as well?

src/vip/data_processor/db/postgres.clj
tie-rack: You can't trim `nil`. But, yes, the only values coming out should be valid at this point.

src/vip/data_processor/db/postgres.clj
tie-rack: The transaction in `save-election-id!` is there because there must be one to head off the (remote but real) possibility of two imports for the same election getting to this point at the same time. Otherwise, like you guess, there's no harm done dying in between these calls, as the import would be re-run, and able to use the row in election_approvals.

src/vip/data_processor/db/postgres.clj
tank157: Ok, cool, I won't worry about it then, just trying to think a bit like a Chaos Human here.

tank157:
I wouldn't mind seeing some tests that exercise the database some more, especially since there's some complexity in the interplay of the `save-election-id!` function, but if that's not something we've been doing I can understand. Otherwise, LGTM

PR #102, opened 2015-10-26
Update missing required field error severity

tie-rack:
The default error severity for most missing required fields was too severe.

* Use a severity level as the value for a `:required` key in a column definition instead of simply `true`
* Update the column definitions to be _mostly_ `:critical` for missing values
* Keep missing fields in `sources` and `elections` at `:fatal`

Pivotal story: [103449132](https://www.pivotaltracker.com/story/show/103449132)


ericnormand:
LGTM

PR #103, opened 2015-10-29
Make the processor less CPU hungry

tie-rack:
`(while true)` churns at ~100% CPU usage, while trying to deref the SQS processing future does not. That future should never deref, so it's safe.

And, if the future throws an exception, the processor will now shut down (whereas before it would just sit there no longer doing work). That's nice.


cap10morgan:
Nice!

ericnormand:
LGTM

PR #107, opened 2015-11-23
Insert all the data into the database

tie-rack:
When inserting multiple rows into a database with korma, the keys of the first row determine which columns get values. So if a later row has a key the first one doesn't, the associated value will not be inserted into the database! This is very bad!

This is only an issue with XML importing, since during a CSV import each row (by virtue of how they're constructed) will share all keys with every other row.

This patch updates `bulk-import` to make all rows share all keys. Only the first row could be updated, but I'd like to protect myself if korma started using, say, the last row to figure out which columns to bother with.

No pivotal story. Found during other work.

### DIFF
--- a/src/vip/data_processor/db/util.clj
+++ b/src/vip/data_processor/db/util.clj
@@ -105,20 +105,32 @@
                            ["Duplicate id"]))
                ctx (set/union existing-ids local-dupe-ids))))))
 
+(defn hydrate-row [ks row]
+  (reduce (fn [row k]
+            (if (contains? row k)
+              row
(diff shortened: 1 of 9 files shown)


cap10morgan:
Ugh, korma...

cap10morgan:
Sad that this is necessary, but the solution looks good to me.

Is there anything worth pushing upstream to the korma folks re: this?

tie-rack:
I'll open an issue on Korma. Later. :)

cap10morgan:
:+1:

PR #108, opened 2015-11-23
Non-mail precincts need a polling location

tie-rack:
Adds a warning for every precinct that are not "mail_only" and are not referenced in the `precinct_polling_locations` table.

Pivotal story: [105526462](https://www.pivotaltracker.com/story/show/105526462)

Note: the error text is a front-end concern and will be addressed in Metis.

### DIFF
--- a/test-resources/csv/missing-polling-locations/precinct.txt
+++ b/test-resources/csv/missing-polling-locations/precinct.txt
@@ -2,3 +2,4 @@ name,number,locality_id,ward,mail_only,ballot_style_image_url,id
 Mail Only Precinct,0034,70123,,yes,,1
 Has Polling Locations,0020,70123,,,,2
 Not Mail Only and No Polling Locations,0021,70123,,,,3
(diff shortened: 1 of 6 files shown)


test-resources/csv/missing-polling-locations/precinct.txt
```diff
@@ -0,0 +1,4 @@
+name,number,locality_id,ward,mail_only,ballot_style_image_url,id
+Mail Only Precinct,0034,70123,,yes,,1
+Has Polling Locations,0020,70123,,,,2
+Not Mail Only and No Polling Locations,0021,70123,,,,3
```
cap10morgan: Does omitting the `mail_only` value result in `0` or `NULL` in the DB? If the latter, does including an explicit `no` (or similar) cause it to put `0` into the DB? Should we add test data for the uncovered case if there is one here?

tie-rack: It results in a `NULL`. I'll add a `no` testcase for coverage. Good call.

cap10morgan:
LGTM

PR #109, opened 2015-11-30
Specify v3.0 data specifications as such

tie-rack:
In preparation for coming 5.0 work, explicitly name the 3.0 `data-specs` and `file-dependencies` so that it's clear they are for 3.0.

Additionally, I've changed any function that implicitly used `data-specs` to take an argument or use the `:data-specs` key on the processing context, so that when a 5.0 specification exists here, the functions will continue to work.

Pivotal story: [107856956](https://www.pivotaltracker.com/story/show/107856956)


tie-rack:
Those `TODO`s will have stories to address them.

ericnormand:
LGTM

tie-rack:
Closing, but this work has been merged into #111.

PR #112, opened 2015-12-02
Sub-namespace xml output namespaces for v3.0.

ericnormand:
Segregate the xml output namespaces into a subnamespace to make room for many, many, many new versions.

This is mostly just moving code into new namespaces (renaming files, moving a few things, and renaming usage). There is also a little bit of logic change that we noticed that had to do with outputting a `nil` where we should have defaulted to the table name (in `vip/data_processor/db/util.clj`).

[Pivotal Card](https://www.pivotaltracker.com/story/show/107858778)


7maples:
LGTM

PR #113, opened 2015-12-11
Namespace pipeline fns

tie-rack:
Not just making 3.0-specific validations explicitly named such, but also taking a couple that _are_ and making them generic.

Pivotal story: [107859888](https://www.pivotaltracker.com/story/show/107859888)


ericnormand:
LGTM

PR #115, opened 2016-01-20
Determine XML import version and branch

tie-rack:
If it's a supported version (just 3.0 for now), load the XML and proceed as usual. Otherwise, stop.


danielglauser:
LGTM

PR #116, opened 2016-01-21
Branch on 5.0 csv import

tie-rack:
Read spec version from the source.txt file (assuming 3.0 if there's no version—per the story and per the 3.0 CSV not reporting its version), and branch based on the result.

Stops processing on unsupported spec version.

Also, normalize naming between XML and CSV version checking.

Pivotal story: [111607092](https://www.pivotaltracker.com/story/show/111607092)

### DIFF
--- a/src/vip/data_processor/validation/csv.clj
+++ b/src/vip/data_processor/validation/csv.clj
@@ -39,6 +39,15 @@
        (filter #(= filename (clojure.string/lower-case (.getName %))))
        first))
 
+(defn read-one-line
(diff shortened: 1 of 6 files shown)


tie-rack:
@danielglauser the XML and CSV checks are still using different names! XML is now `schema-version` and CSV is `spec-version`. They should be aligned.

danielglauser:
@tie-rack Curses! Which name would you like to win?

tie-rack:
I think `spec-version`.

tie-rack:
LGTM!

PR #117, opened 2016-02-08
Add xml tree index

tie-rack:
Add an indexes on the XML import table to improve queries.

Also, because the table isn't limited to 5.0 feeds, renames `v5_0_xml_values` to `xml_tree_values`.

Why just one index? Well, it appears we only need one. Naturally, a B-Tree index on `(results_id, path)` is used by queries like `SELECT * FROM xml_tree_values WHERE results_id = ? AND path = ?;` directly.

We assumed that adding a GIST index on `path` would help queries like `SELECT * FROM xml_tree_values WHERE results_id = ? AND path ~ ?;`, however a GIST index on that column cannot be part of a multi-column index that includes the `results_id` (because GIST doesn't work on integers). By providing a GIST index on `path`, the query planner uses it, though without being able to constrain the index to to `results_id` first.

On the other hand, without that index, Postgres can use the index on `results_id` to produce a much constrained working set, then filter them for the matches on `path`, which (based on some benchmarking) works better for us. Thus, _not_ indexing it performs better (for this use case, with our kind of data).

Pivotal story: [112938309](https://www.pivotaltracker.com/story/show/112938309)

### DIFF
--- a/resources/migrations/20160126-create-5-0-xml-import-table.down.sql
+++ b/resources/migrations/20160126-create-5-0-xml-import-table.down.sql
@@ -1,1 +1,0 @@
-DROP TABLE v5_0_xml_values;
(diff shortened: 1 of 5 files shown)


dcohen:
LGTM!

PR #118, opened 2016-02-09
Add xml_tree_validations table

tie-rack:
Pivotal story: [112939295](https://www.pivotaltracker.com/story/show/112939295)


tie-rack:
This PR _is_ against `import-5.0-xml-into-postgres`

dcohen:
Yup, so it is! Looks good.

PR #120, opened 2016-02-16
Validate 5.0 emails

tie-rack:
Emails should look nominally like emails. Now we add errors for 5.0 feeds when they don't.

The big important change, however, is that tests that use a Postgres database are both available and run on the Buildkite server.

[This build](https://buildkite.com/the-voting-information-project/qa-engine/builds/376) includes an extra build step, "Postgres tests", which runs the `script/postgres-tests` script. That script uses `docker-compose` to run a Postgres container for tests marked with the `^:postgres` metadata.

Outside of `docker-compose`, running the postgres tests requires you to specify that you want to run them. `lein test` will not, by default. You must either run `lein test :postgres` to run just the Postgres tests, or `lein test :all` to run all the tests. You will also want a test-resources/config.edn file with the `:postgres` section configured to use a test database.

Future tests that use Postgres should be in a namespace just for tests that use Postgres (the nature of clojure.test fixtures makes it better to separate them by namespace). Test namespaces that need Postgres should have calls to `(use-fixtures :once setup-postgres)` to initialize Postgres and `(use-fixtures :each with-clean-postgres)` to clear the test database between tests. See `test/vip/data_processor/validation/xml_test_with_postgres.clj` as an example.

Pivotal story: [112940827](https://www.pivotaltracker.com/story/show/112940827)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,3 +9,6 @@ script
 .gitignore
 .lein-*
 .nrepl-port
+test-resources/config.edn
+docker-compose-postgres-tests.yml
+docker-compose.yml
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ pom.xml.asc
 /.nrepl-port
 /dev-resources/config.edn
 data-processor@.service
+/test-resources/config.edn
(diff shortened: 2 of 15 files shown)


dcohen:
Tests pass, LGTM.

PR #121, opened 2016-02-18
Validate uniqueness of ids in imported 5.0 feeds

sduckett:
Any duplicate IDs will be included under the key `:fatal` in the processing context. We also fixed a problem where Postgres tests in multiple namespaces would not run.

[Pivotal Card](https://www.pivotaltracker.com/story/show/112941181)


dcohen:
LGTM.

tie-rack:
One more commit, @dcohen

PR #123, opened 2016-02-23
Validate presence of ids

tie-rack:
Validate that every element that needs an id in a 5.0 feed has one.

This PR includes the commits from #122, and really only encompasses 6758cd8.

[Pivotal story](https://www.pivotaltracker.com/story/show/113392855)


dcohen:
LGTM

PR #124, opened 2016-02-25
Load errors only when errors exist

sduckett:
If there are no errors, we'll generate bad SQL and the import process will fail.


tie-rack:
LGTM. Thank you!

PR #126, opened 2016-03-03
Validate candidate elements

cap10morgan:
@tie-rack I think I have a working BallotName validator here.

(Re-opened against the correct branch)

[Pivotal card](https://www.pivotaltracker.com/story/show/113397267)

### DIFF
--- a/src/vip/data_processor/validation/v5.clj
+++ b/src/vip/data_processor/validation/v5.clj
@@ -4,7 +4,8 @@
             [vip.data-processor.validation.v5.id :as id]))
 
 (def validations
-  [candidate/validate-post-election-statuses
+  [candidate/validate-no-missing-ballot-names
+   candidate/validate-post-election-statuses
    email/validate-emails
    id/validate-unique-ids
    id/validate-no-missing-ids])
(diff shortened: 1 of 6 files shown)


tie-rack:
Oh this is pretty close!

The BallotName element, however, doesn't contain simple text content, but rather [InternationalizedText](https://github.com/votinginfoproject/vip-specification/blob/v5.0-release/vip_spec.xsd#L370-L399). So a one would look like [this](https://github.com/votinginfoproject/vip-specification/blob/v5.0-release/sample_feed.xml#L920-L922):

```xml
    <BallotName>
      <Text language="en">Joseph R. Biden</Text>
    </BallotName>
```

Additionally, the `.0` in the generated paths that you're checking for assume that BallotName elements are always the first child of Candidates.

cap10morgan:
Ah, OK. I'll get that fixed up.

cap10morgan:
This should be ready for review again. I am surprised, though, that build 90646d43521b9ef0d0c85d44f89b9b82026bcdc0 passed. After fixing those tests locally, the build failed (and rightly so). Commit f899353 fixed it.

tie-rack:
Just add `candidate/validate-no-missing-ballot-names` to `vip.data-processor.validation.v5/validations` and this should be good to go!

cap10morgan:
I did that already, I think?

tie-rack:
Sorry, I mean `validate-pre-election-statuses`!

tie-rack:
LGTM!

PR #127, opened 2016-03-07
Validate precincts

tie-rack:
Validate that `Precinct` elements have `Name` and `LocalityId` children.

Pivotal story: [113421735](https://www.pivotaltracker.com/story/show/113421735)

Additionally, now that "run a query and add an error for every path returned" is a pattern, pull it out and reduce the amount of repeated structure that we started littering these namespaces with.

### DIFF
--- a/src/vip/data_processor/validation/v5/util.clj
+++ b/src/vip/data_processor/validation/v5/util.clj
@@ -11,7 +11,7 @@
 (defn build-xml-tree-value-query-validator
   "Generate a validator that adds a validation error for every path in
   the results of the query. The params-fn must be a fn of one argument (the
-  context), which reutrns a vector of params for the query."
+  context) which returns a vector of params for the query."
   [severity scope error-type error-data query params-fn]
   (fn [{:keys [import-id] :as ctx}]
     (let [missing-paths (korma/exec-raw
--- a/src/vip/data_processor/validation/v5/util.clj
+++ b/src/vip/data_processor/validation/v5/util.clj
@@ -1,0 +1,26 @@
+(ns vip.data-processor.validation.v5.util
+  (:require [korma.core :as korma]
+            [vip.data-processor.db.postgres :as postgres]))
+
+(defn two-import-ids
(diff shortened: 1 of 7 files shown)


src/vip/data_processor/validation/v5/util.clj
```diff
@@ -0,0 +1,26 @@
...
+(defn two-import-ids
+  "A common params-fn for build-xml-tree-value-query-validator that
+  returns a 2-element vector of the import-id of the context twice."
+  [{:keys [import-id]}]
+  [import-id import-id])
+
+(defn build-xml-tree-value-query-validator
+  "Generate a validator that adds a validation error for every path in
+  the results of the query. The params-fn must be a fn of one argument (the
+  context), which reutrns a vector of params for the query."
```
cap10morgan: s/reutrns/returns/

src/vip/data_processor/validation/v5/util.clj
```diff
@@ -0,0 +1,26 @@
...
+    (let [missing-paths (korma/exec-raw
+                         (:conn postgres/xml-tree-values)
+                         [query (params-fn ctx)]
+                         :results)]
+      (->> missing-paths
+           (map :path)
+           (reduce (fn [ctx path]
+                     (update-in ctx [severity scope (.getValue path) error-type]
+                                conj error-data))
+                   ctx)))))
```
cap10morgan: This is a nice improvement, and something I was thinking about too. Though it still seems like there's a lot of boilerplate in the query string itself. Could we pull any of that into this fn (possibly in lower-arity version so it could be overridden in exceptional cases like the missing-id query)?

tie-rack: Knowing a little about the validations we still have coming, it looks like there's a few patterns of what these queries will look like, and none will have enough of a plurality to make it seem a good candidate for a default case. I think this might be a good thing to wait and see on.

cap10morgan:
Fair enough. In that case, LGTM!

PR #128, opened 2016-03-07
Add RetentionContest -> CandidateId validation

cap10morgan:
Validates that all RetentionContest elements have CandidateId child elements.

[Pivotal card](https://www.pivotaltracker.com/story/show/113421645)

### DIFF
--- a/src/vip/data_processor/validation/v5/retention_contest.clj
+++ b/src/vip/data_processor/validation/v5/retention_contest.clj
@@ -1,7 +1,5 @@
 (ns vip.data-processor.validation.v5.retention-contest
-  (:require [korma.core :as korma]
-            [vip.data-processor.db.postgres :as postgres]
-            [vip.data-processor.validation.v5.util :as util]))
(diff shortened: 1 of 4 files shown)


src/vip/data_processor/validation/v5/retention_contest.clj
```diff
@@ -0,0 +1,16 @@
+(ns vip.data-processor.validation.v5.retention-contest
+  (:require [korma.core :as korma]
+            [vip.data-processor.db.postgres :as postgres]
+            [vip.data-processor.validation.v5.util :as util]))
```
tie-rack: You don't need to require korma or postgres here.

tie-rack:
And add the validation to the list in v5.clj.

src/vip/data_processor/validation/v5/retention_contest.clj
cap10morgan: Ah, that's now true! (I started this before you created `build-xml-tree-value-query-validator`.)

tie-rack:
LGTM

PR #129, opened 2016-03-07
Validate States

tie-rack:
States need Names.

Pivotal story: [113421593](https://www.pivotaltracker.com/story/show/113421593)

### DIFF
--- a/test/vip/data_processor/validation/v5/state_test.clj
+++ b/test/vip/data_processor/validation/v5/state_test.clj
@@ -16,7 +16,7 @@
                         state/validate-no-missing-names]}
         out-ctx (pipeline/run-pipeline ctx)]
     (testing "missing Names are flagged"
-      (is (get-in out-ctx [:errors :states "VipObject.0.State.0.Name" :missing]))
-      (is (get-in out-ctx [:errors :states "VipObject.0.State.1.Name" :missing])))
+      (is (get-in out-ctx [:fatal :states "VipObject.0.State.0.Name" :missing]))
+      (is (get-in out-ctx [:fatal :states "VipObject.0.State.1.Name" :missing])))
     (testing "doesn't for those that aren't"
-      (is (not (get-in out-ctx [:errors :states "VipObject.0.State.2.Name" :missing]))))))
+      (is (not (get-in out-ctx [:fatal :states "VipObject.0.State.2.Name" :missing]))))))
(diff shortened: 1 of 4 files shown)


test/vip/data_processor/validation/v5/state_test.clj
```diff
@@ -0,0 +1,22 @@
...
+  (let [ctx {:input (xml-input "v5-states.xml")
+             :pipeline [psql/start-run
+                        xml/load-xml-ltree
+                        state/validate-no-missing-names]}
+        out-ctx (pipeline/run-pipeline ctx)]
+    (testing "missing Names are flagged"
+      (is (get-in out-ctx [:errors :states "VipObject.0.State.0.Name" :missing]))
+      (is (get-in out-ctx [:errors :states "VipObject.0.State.1.Name" :missing])))
+    (testing "doesn't for those that aren't"
+      (is (not (get-in out-ctx [:errors :states "VipObject.0.State.2.Name" :missing]))))))
```
cap10morgan: The card says these should be `:fatal` errors. Isn't that a top-level `:fatal` key instead of `:errors`?

test/vip/data_processor/validation/v5/state_test.clj
```diff
@@ -0,0 +1,22 @@
...
+  (let [ctx {:input (xml-input "v5-states.xml")
+             :pipeline [psql/start-run
+                        xml/load-xml-ltree
+                        state/validate-no-missing-names]}
+        out-ctx (pipeline/run-pipeline ctx)]
+    (testing "missing Names are flagged"
+      (is (get-in out-ctx [:fatal :states "VipObject.0.State.0.Name" :missing]))
+      (is (get-in out-ctx [:fatal :states "VipObject.0.State.1.Name" :missing])))
+    (testing "doesn't for those that aren't"
+      (is (not (get-in out-ctx [:fatal :states "VipObject.0.State.2.Name" :missing]))))))
```
cap10morgan: I've started using `(is (not (:fatal out-ctx)))` for tests like this to make it more likely that it will fail if some error is happening. But maybe this is actually another place we should use `assert-no-problems` instead? I like broad tests when it's using `(is (not ...))` so that small inconsistencies don't slip through as false positives.

tie-rack: In this case, there _are_ fatal errors on that context. But `(assert-no-problems out-ctx [:states "VipObject.0.State.2.Name"])` would be broader.

cap10morgan: Must there be other fatal errors? Do they make it a better test?

tie-rack: It's not necessary. It's just that a single XML file is being used to test all cases.

cap10morgan:
LGTM

PR #130, opened 2016-03-08
v5 full good run test

tie-rack:
Add a test that exercises all the v5 validations on a good XML feed. Updates the v5_sample_feed.xml file to include data to check tests that would otherwise not have data to work on.


cap10morgan:
Is the idea that once this lands, we should add new valid elements (if they aren't already there) to v5_sample_feed.xml in subsequent validation PRs to keep this in sync?

tie-rack:
@cap10morgan yeah, we should strive to keep it as full of testable elements as reasonable.

cap10morgan:
LGTM

PR #131, opened 2016-03-08
Add Source element validations

cap10morgan:
Opening a PR on this now (the card calls for additional validations that aren't in here yet) to get early feedback since there's just enough novelty in this first validation to warrant it.

[Pivotal card](https://www.pivotaltracker.com/story/show/113421549)

### DIFF
--- a/src/vip/data_processor/validation/v5/source.clj
+++ b/src/vip/data_processor/validation/v5/source.clj
@@ -45,3 +45,17 @@
       ctx
       (update-in ctx [:fatal :source path :missing]
                  conj :missing-date-time))))
+
+(defn validate-vip-id [{:keys [import-id] :as ctx}]
+  (let [path "VipObject.0.Source.*{1}.VipId.*{1}"
+        source-vip-id (-> (korma/select postgres/xml-tree-values
+                                        (korma/where {:results_id import-id})
+                                        (korma/where
+                                         (postgres/ltree-match
+                                          postgres/xml-tree-values :path path)))
+                          first
+                          :value)]
+    (if source-vip-id
+      ctx
+      (update-in ctx [:fatal :source path :missing]
+                 conj :missing-vip-id))))
--- a/src/vip/data_processor/validation/v5/source.clj
+++ b/src/vip/data_processor/validation/v5/source.clj
@@ -1,0 +1,19 @@
+(ns vip.data-processor.validation.v5.source
+  (:require [korma.core :as korma]
+            [vip.data-processor.db.postgres :as postgres]))
+
+(defn validate-one-source [{:keys [import-id] :as ctx}]
+  (let [result (korma/exec-raw
+                (:conn postgres/xml-tree-values)
+                ["SELECT COUNT(DISTINCT subltree(path, 0, 4)) AS source_count
+                  FROM xml_tree_values
+                  WHERE results_id = ?
+                  AND path ~ 'VipObject.0.Source.*'" [import-id]]
+                :results)
+        source-count (-> result
+                         first
+                         :source_count)]
+    (if (> source-count 1)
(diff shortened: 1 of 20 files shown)


src/vip/data_processor/validation/v5/source.clj
```diff
@@ -0,0 +1,34 @@
...
+        source-count (-> result
+                         first
+                         :source_count)]
+    (if (> source-count 1)
+      (update-in ctx [:fatal :source "VipObject.0.Source" :count]
+                 conj :more-than-one)
+      ctx)))
+
+(defn validate-source-name [{:keys [import-id] :as ctx}]
+  (let [path "VipObject.0.Source.0.Name.0"
```
tie-rack: This path assumes that the Source is the first child of VipObject, and that Name is the first child of that Source. Neither should be counted on.

cap10morgan: Right, right. It's going to take me a few more screw-ups before I get that straight. It *feels* so hierarchical...

cap10morgan:
OK, I think this is ready for final review.

src/vip/data_processor/validation/v5/source.clj
```diff
@@ -0,0 +1,76 @@
...
+                :results)
+        source-count (-> result
+                         first
+                         :source_count)]
+    (if (> source-count 1)
+      (update-in ctx [:fatal :source "VipObject.0.Source" :count]
+                 conj :more-than-one)
+      ctx)))
+
+(defn validate-name [{:keys [import-id] :as ctx}]
```
tie-rack: The rest of the "validate this exists" validations have used `build-xml-tree-value-query-validator`. Why not these?

test/vip/data_processor/validation/v5/source_test.clj
```diff
@@ -0,0 +1,123 @@
...
+                          v5.source/validate-one-source]}
+          out-ctx (pipeline/run-pipeline ctx)]
+      (is (get-in out-ctx [:fatal :source "VipObject.0.Source" :count]))))
+  (testing "one and only one Source element is OK"
+    (let [ctx {:input (xml-input "v5-one-source.xml")
+               :pipeline [psql/start-run
+                          xml/load-xml-ltree
+                          v5.source/validate-one-source]}
+          out-ctx (pipeline/run-pipeline ctx)]
+      (is (not (get-in out-ctx [:fatal :source "VipObject.0.Source" :count]))))))
```
tie-rack: For tests that expect no validation errors of any kind, like this one, may be better served by `(assert-no-problems out-ctx [])`.

src/vip/data_processor/validation/v5/source.clj
cap10morgan: These seemed simpler since we only care whether or not we could find something at that path or not (versus needing to compare many different expected paths to actual paths). But if you'd rather see these use `build-xml-tree-value-query-validator`, I'd be happy to convert them to that.

test/vip/data_processor/validation/v5/source_test.clj
cap10morgan: I couldn't agree more! ;)

src/vip/data_processor/validation/v5/source.clj
tie-rack: Using a `build-xml-tree-value-query-validator`-style validator would have the benefit of providing a real path in the error instead of an lquery, which may make eventual error messaging better.

cap10morgan: True. OK, I'll change them.

cap10morgan:
Anything else to do on this?

tie-rack:
Assuming Buildkite approves, I do.

PR #134, opened 2016-03-09
Add HoursOpen time format validation

cap10morgan:
...for StartTime and EndTime descendants.

I also moved some boilerplate korma code I'd repeated a few times into the util namespace. If you like it, I'll do a subsequent PR to convert other uses of that pattern to use the util fn instead.

[Pivotal card](https://www.pivotaltracker.com/story/show/114350619)


src/vip/data_processor/validation/v5/util.clj
```diff
@@ -24,3 +24,17 @@
...
                                 conj error-data))
                    ctx)))))
+
+(defn select-path
+  "Returns the xml-tree-values whose results_id matches the given `import-id`
+  and that match an ltree query on `path`. Returns a korma results collection.
+
+  Example: (select-path 1 \"VipObject.0.Source.id\") will run a query like:
+           SELECT * FROM xml_tree_values WHERE results_id = 1
+           AND path ~ 'VipObject.0.Source.id'"
```
tie-rack: For queries to a specific path, like `VipObject.0.Source.id`, you'd want to use `=`, not `~`, so the query could be as simple as: `(korma/where {:results_id import-id :path path})`. I see that's not how it's being used in this PR, but the comment, the name of the fn, and the name of the second argument all suggest the wrong thing to do.

This may be better named `select-lquery` (since we're using the lquery match operator, `~`) with the second arg named `lquery`.

src/vip/data_processor/validation/v5/hours_open.clj
```diff
@@ -0,0 +1,22 @@
...
+  (re-matches
+   #"(?:(?:[01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]|(?:24:00:00))(?:Z|[+-](?:(?:0[0-9]|1[0-3]):[0-5][0-9]|14:00))"
+   time))
+
+(defn validate-times [{:keys [import-id] :as ctx}]
+  (let [hours-open-path "VipObject.0.HoursOpen.*{1}.Schedule.*{1}.Hours.*{1}"
+        start-times (util/select-path import-id
+                                      (str hours-open-path ".StartTime.*{1}"))
+        end-times (util/select-path import-id
+                                    (str hours-open-path ".EndTime.*{1}"))
```
tie-rack: It looks like you can get both start and end times with a single lquery: `VipObject.0.HoursOpen.*{1}.Schedule.*{1}.Hours.*{1}.StartTime|EndTime.*{1}`. See the lquery section of the [ltree docs](http://www.postgresql.org/docs/9.1/static/ltree.html).

cap10morgan:
@tie-rack Anything else?

tie-rack:
That's it. Resolve those merge conflicts and you're LGTM.

src/vip/data_processor/validation/v5/hours_open.clj
```diff
@@ -0,0 +1,19 @@
+(ns vip.data-processor.validation.v5.hours-open
+  (:require [vip.data-processor.validation.v5.util :as util]))
+
+(defn valid-time-with-zone? [time]
+  (re-matches
+   #"(?:(?:[01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]|(?:24:00:00))(?:Z|[+-](?:(?:0[0-9]|1[0-3]):[0-5][0-9]|14:00))"
```
tie-rack: Should this have anchors (`\A` and `\z`) to ensure we're matching the entire string?

cap10morgan: Yeah, that seems like a good idea.

PR #137, opened 2016-03-17
type->lqueries

tie-rack:
Given a type for an XSD, generate lqueries to find all elements of that type. For help with numerous validations to come, and for future retrofitting of validations gone by.

### DIFF
--- a/src/vip/data_processor/validation/xml/xpath.clj
+++ b/src/vip/data_processor/validation/xml/xpath.clj
@@ -1,0 +1,106 @@
+(ns vip.data-processor.validation.xml.xpath
+  (:import [javax.xml.parsers DocumentBuilder DocumentBuilderFactory]
+           [javax.xml.xpath XPathFactory XPathConstants])
+  (:require [clojure.java.io :as io]))
+
+(defn spec-resource [version]
+  (let [path (str "specs/vip_spec_v" version ".xsd")]
+    (io/resource path)))
+
+(def doc-builder-factory (doto (DocumentBuilderFactory/newInstance)
+                           (.setNamespaceAware true)))
+(def doc-builder (.newDocumentBuilder doc-builder-factory))
+
+(defn parse-spec-by-version [version]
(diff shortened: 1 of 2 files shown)


tie-rack:
Buildkite doesn't like `>` in branch names.

src/vip/data_processor/validation/xml/xpath.clj
```diff
@@ -0,0 +1,106 @@
+(ns vip.data-processor.validation.xml.xpath
```
cap10morgan: `xpath` seems like an implementation detail. Should we call this namespace `xml.spec`? A lot of the `spec` references in several of the fn names could potentially go away then.

src/vip/data_processor/validation/xml/xpath.clj
```diff
@@ -0,0 +1,106 @@
...
+(defn spec-resource [version]
+  (let [path (str "specs/vip_spec_v" version ".xsd")]
+    (io/resource path)))
+
+(def doc-builder-factory (doto (DocumentBuilderFactory/newInstance)
+                           (.setNamespaceAware true)))
+(def doc-builder (.newDocumentBuilder doc-builder-factory))
+
+(defn parse-spec-by-version [version]
+  (with-open [is (io/input-stream (spec-resource version))]
```
cap10morgan: Does this work in an uberjar?

src/vip/data_processor/validation/xml/xpath.clj
```diff
@@ -0,0 +1,106 @@
...
+    (-> type-query-expr
+        (query-spec version)
+        nodeset->seq)))
+
+(declare paths-for-type)
+
+(defn paths-for-element [element version]
+  (let [ancestors (ancestry element)]
+    (reduce (fn [paths ancestor]
+              (condp = (.getTagName ancestor)
```
cap10morgan: Since both of your cases are string literals, would this be better as `(case (.getTagName ancestor) ...)`?

PR #138, opened 2016-03-17
type->lqueries

tie-rack:
Given a type for an XSD, generate lqueries to find all elements of that type.

For help with validations to come, and retrofitting validations gone by.


cap10morgan:
See comments on https://github.com/votinginfoproject/data-processor/pull/137

tie-rack:
@cap10morgan Suggestions acted on. And confirmed that it works in an uberjar.

cap10morgan:
LGTM

PR #139, opened 2016-03-18
Faster xml type queries

tie-rack:
* Adds a column to `xml_tree_values` called `simple_path`, which is like `path`, but without the indexes (and thus is not good for _identifying_ data, but good for _querying_ data in ways we want to).
* Replaces `type->lqueries` with `type->simple-paths`, because equality checks are faster than `~`.
* Updates current validations that could trivially use `simple_path` to do so.


cap10morgan:
LGTM

PR #142, opened 2016-03-22
Find *all* the simple_paths for a type

tie-rack:
type->simple-paths was previously not following chains of ancestry from types being extended into other types. That is fixed, now, along with some refactoring that simplified matters.

Meanwhile, the XPath querying became Namespace-aware, so that we could query for "xs:extension" elements, needed for this to work.

Phew.


cap10morgan:
LGTM

PR #143, opened 2016-03-23
Validate Locality elements

cap10morgan:
This isn't quite done yet, but I wanted to get it up for review of my general approach to integrating the new type-lookup code and simple paths.

Instead of assuming one element named `Locality` that is a child of `VipObject`, this looks up all elements of the type `Locality` from the schema and runs its validations on them.

[Pivotal card](https://www.pivotaltracker.com/story/show/114351679)


tie-rack:
This approach looks good to me!

cap10morgan:
OK, this should be ready for final review now.

I think there are some things to extract into the util namespace once one of us does the next similar validation. But there's nothing driving that in this PR quite yet.

tie-rack:
Yep. LGTM.

PR #144, opened 2016-03-23
Validate internationalized text

cap10morgan:
The big kahuna. [Pivotal card](https://www.pivotaltracker.com/story/show/114351529)

### DIFF
--- a/src/vip/data_processor/validation/v5.clj
+++ b/src/vip/data_processor/validation/v5.clj
@@ -10,7 +10,8 @@
             [vip.data-processor.validation.v5.hours-open :as hours-open]
             [vip.data-processor.validation.v5.election-administration :as election-admin]
             [vip.data-processor.validation.v5.election :as election]
-            [vip.data-processor.validation.v5.locality :as locality]))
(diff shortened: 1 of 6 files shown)


tie-rack:
I had been thinking about writing `build-no-missing-validators` the way you've done here to handle the last argument to `subltree`. I'm glad you did it!

Since it looks like the `reduce` over the validators from `build-no-missing-validators` is going to copied all over the place, do you think it would make sense for another helper? Like

```clj
(defn validate-no-missing-validators
  [schema-type element]
  (fn [{:keys [import-id] :as ctx}]
    (let [validators (util/build-no-missing-validators
                      schema-type
                      element
                      import-id)]
      (reduce (fn [ctx validator] (validator ctx)) ctx validators))))
```

And then replace, for example, the definition of `validate-no-missing-texts` with:

```clj
(def validate-no-missing-texts
  (validate-no-missing-validators :internationalized-text :text))
```

And similarly throughout?

cap10morgan:
I think that's a smashing idea. What do you think of this implementation of it?

tie-rack:
Looks great. :LGTM:

PR #145, opened 2016-03-25
let election_type be longer than 20 characters

tie-rack:
Some recent feeds have had longer `election_type`s than could be stored. So let's remove the very constricting type with a more liberating one.

There are no other limited `VARCHAR` types in the tables. I checked.


cap10morgan:
Assuming the build failure is the build config's fault, this LGTM.

PR #146, opened 2016-03-29
Validate Office element

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/114352615)

While writing this I noticed another opportunity for some code centralization, which drove out the `vip.data-processor.validation.v5.util/validate-enum-elements` fn.

After making the Office -> Term -> Type validation use it, I also adapted the other validations already in the target branch to use it too.

### DIFF
--- a/src/vip/data_processor/validation/v5/util.clj
+++ b/src/vip/data_processor/validation/v5/util.clj
@@ -100,27 +100,35 @@
                                                   import-id)]
       (reduce (fn [ctx validator] (validator ctx)) ctx validators))))
 
-(defn validate-enum-elements
-  "Returns a fn that takes a validation `ctx` and runs an enumerated values
-  validation on all elements of `schema-type` in the `ctx`'s import, checking
-  that all of their values are one of those enumerated as valid by the schema."
-  [schema-type]
+(defn elements-for-simple-path
+  "Returns all elements in `import-id` whose `simple-path` matches the given
+  arg."
+  [import-id simple-path]
+  (korma/select postgres/xml-tree-values
+    (korma/where
+     {:results_id import-id
+      :simple_path (postgres/path->ltree simple-path)})))
+
+(defn nth-to-last-path-element
+  "Returns the ltree path element `n` segments from the end."
+  [path n]
+  (-> path
+      (str/split #"\.")
+      reverse
+      (nth n)))
+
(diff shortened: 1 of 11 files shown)


src/vip/data_processor/validation/v5/util.clj
```diff
@@ -131,3 +139,15 @@
...
+(defn validate-enum-elements
+  "Returns a fn that takes a validation context and runs an enumerated values
+  validation on all elements of `schema-type` in the context's import, checking
+  that all of their values are one of those enumerated as valid by the schema.
+
+  See `validate-elements` for more details."
+  [schema-type]
+  (let [xml-schema-type (keyword->xml-name schema-type)
+        valid-values (spec/enumeration-values xml-schema-type "5.0")
+        valid? #(valid-values %)]
```
tie-rack: `#(x %)` is an anonymous layer of indirection over `x`. Why bother?

cap10morgan: Yeah, I thought about that too. It seemed much clearer for those reading it later? But I guess either way it's relying on the magical set-in-fn-position behavior in Clojure. I'll take it out if you'd prefer just binding the set to `valid?`.

src/vip/data_processor/validation/v5/util.clj
```diff
@@ -100,27 +100,35 @@
...
+
+(defn nth-to-last-path-element
+  "Returns the ltree path element `n` segments from the end."
+  [path n]
+  (-> path
+      (str/split #"\.")
+      reverse
+      (nth n)))
+
+(defn validate-elements
```
tie-rack: I'm guessing this will eventually need to take the error severity as an argument. That all validations that use this so far have used `:errors` is a coincidence.

cap10morgan: Yeah, I should just add that now. I'll do that.

src/vip/data_processor/validation/v5/util.clj
tie-rack: For clarity, I'd take `#(contains? valid-values %)`, which is still assuming that `valid-values` is a set, but it does read pretty well. Or I'd take binding the set to `valid?`. Either way.

tie-rack:
LGTM

PR #147, opened 2016-03-29
Validate OrderedContest element

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/114352989)


tie-rack:
LGTM

PR #148, opened 2016-03-29
Validate Party -> Color

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/114353319)

This is based on the same branch as #146 so I could use the `util/validate-elements` fn from there. But only commit d91587cce4ecd2477eea365944ee0f8221d1e938 and forward are relevant to this PR.


tie-rack:
LGTM!

PR #150, opened 2016-03-31
Update deps

cap10morgan:
Apparently this is how I entertain myself on my day off.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM clojure:lein-2.5.0
+FROM clojure:lein-2.5.3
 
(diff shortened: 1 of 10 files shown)


cap10morgan:
Hmm... the tests all pass locally. I'll have to look into that CI failure.

cap10morgan:
I guess all it needed was another dependency update!

tie-rack:
@cap10morgan This LGTM now (now that I fixed one lingering problem from updating Korma). Take a look at the change I pushed.

cap10morgan:
Ah, good catch. How did you notice that?

PR #151, opened 2016-04-04
The simple street segment validations

tie-rack:
Pivotal story: [114358811](https://www.pivotaltracker.com/story/show/114358811)

Validate the presence of OddEvenBoth, StreetName, City, State, Zip and the format of OddEvenBoth.

And one macro.


cap10morgan:
LGTM

PR #152, opened 2016-04-04
Validate PollingLocation elements

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/114353949)

Note that this PR modifies the 5.0 schema! It's equivalent as far as valid XML files go, but it gives the `LatLng` type a name (also `LatLng`) because our schema-based validation generators currently rely on named types.


tie-rack:
Cool. LGTM.

PR #154, opened 2016-04-05
Validate ballot measure contests

tie-rack:
BallotMeasureContests need a Type and it needs to be a valid Type.

And it turns out the sample feed had a problem in this regard. That is fixed.

Pivotal story: [114348619](https://www.pivotaltracker.com/story/show/114348619)


cap10morgan:
LGTM modulo the answer to the "selection" question in the card.

PR #156, opened 2016-04-22
Upload 5.0 feeds to s3

tie-rack:
Uploads validated 5.0 XML files to s3 after validation.

Also updates the processing pipeline to handle 5.0 feeds. There were TODOs that were finally addressed.

This is deployed to staging for demonstration purposes.

Pivotal story: [116161903](https://www.pivotaltracker.com/story/show/116161903)

### DIFF
--- a/src/vip/data_processor.clj
+++ b/src/vip/data_processor.clj
@@ -26,20 +26,26 @@
    zip/assoc-file
    zip/extracted-contents])
 
+(def v3-validation-pipeline
+  (concat db/validations
+          db.v3-0/validations
+          xml-output/pipeline
+          [psql/insert-validations
+           psql/import-from-sqlite
+           psql/store-stats]))
+
 (def pipeline
   (concat download-pipeline
           [(data-spec/add-data-specs v3-0/data-specs) ; TODO: decide which specs to add based on import
            t/remove-invalid-extensions
            t/xml-csv-branch
            psql/store-public-id
-           psql/store-election-id]
-          db/validations
-          db.v3-0/validations ; TODO: choose extra validations based on import
-          xml-output/pipeline
-          [s3/upload-to-s3]
-          [psql/insert-validations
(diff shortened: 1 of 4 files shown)


src/vip/data_processor.clj
```diff
@@ -26,20 +26,26 @@
...
-          xml-output/pipeline
-          [s3/upload-to-s3]
-          [psql/insert-validations
-           psql/import-from-sqlite
-           psql/store-stats
+           psql/store-election-id
+           (fn [{:keys [spec-version] :as ctx}]
+             (condp = spec-version
+               "3.0" (update ctx :pipeline (partial concat v3-validation-pipeline))
+               ctx))
```
cap10morgan: So this works for 5.0 now (even with the `v3-0/data-specs` TODO still undone above)?

Would it be better to have an explicit "5.0" case and throw an error as the default case?

tie-rack: This does work for 5.0 even with the `v3-0/data-specs`, since none of the 5.0 validations ever look at the attached data-spec. I'll update that, though, nonetheless.

As to your second question: I'm not sure. Bad versions will already be thrown out by this part of the pipeline. And it's only that 3.0 needs extra functions here.

cap10morgan: OK, seems fine then.

cap10morgan:
LGTM

src/vip/data_processor.clj
tie-rack: Spoke too soon on point 1. That will require some more extensive refactoring. There are CSV validation functions that use the `data-spec` before it's known which version an import is.

However, the next thing I'm tackling is 5.0 CSVs, which will provide a good spot to do that work.

cap10morgan: OK, yeah, if it's not breaking anything currently then the 5.0 CSV work sounds like the right context in which to tackle it.

PR #158, opened 2016-05-05
Create zip file in a temp directory

tie-rack:
Fixes the problem where a zip file can carry over between processing runs of feeds for the same election and causing problems when trying to recreate the zip file.

### DIFF
--- a/src/vip/data_processor/s3.clj
+++ b/src/vip/data_processor/s3.clj
@@ -53,8 +53,10 @@
   "Uploads the generated xml file to the specified S3 bucket."
   [ctx]
   (let [zip-name (str (zip-filename ctx) ".zip")
-        zip-file (File. zip-name)
-        zip (ZipFile. zip-name)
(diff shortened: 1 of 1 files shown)


src/vip/data_processor/s3.clj
```diff
@@ -53,8 +53,10 @@
   "Uploads the generated xml file to the specified S3 bucket."
   [ctx]
   (let [zip-name (str (zip-filename ctx) ".zip")
-        zip-file (File. zip-name)
-        zip (ZipFile. zip-name)
+        zip-dir (Files/createTempDirectory tmp-path-prefix
```
tank157: Should there be a step where you delete the temp dir? I'm not clear on lifecycles of temp dirs but I thought it was only restarts that cleared them out and we don't do that often.

tie-rack: Good point. Addressed.

tank157:
LGTM

PR #159, opened 2016-05-13
Create v5 tables

sduckett:
Here are the tables for the 5.0 CSVs as we currently understand the spec.

[Pivotal Card](https://www.pivotaltracker.com/story/show/118433041)

This blocks a [data-spec pull request](https://github.com/votinginfoproject/data-processor/pull/160) and an [import csv pull request](https://github.com/votinginfoproject/data-processor/pull/161)


danielglauser:
@sduckett This seems very spec dependent. Could you please post a link to the spec?

sduckett:
@danielglauser This is based on the [XML spec](https://github.com/votinginfoproject/vip-specification/blob/vip511/vip_spec.xsd), and a set of discussions; we don't have a formal 5.0 CSV spec at the moment unfortunately.

resources/migrations/20160509-create-5.0-csv-tables.up.sql
```diff
@@ -0,0 +1,285 @@
...
+                              PRIMARY KEY (results_id, id, schedule_id));
+
+CREATE TABLE v5_0_schedules (id TEXT NOT NULL,
+                             results_id BIGINT REFERENCES results (id) NOT NULL,
+                             PRIMARY KEY (results_id, id),
+                             start_time TEXT,
+                             end_time TEXT,
+                             start_time2 TEXT,
+                             end_time2 TEXT,
+                             is_only_by_appointment TEXT,
```
danielglauser: Why do all of the `is_...` columns read as boolean but are TEXT attributes?

sduckett: To answer, I'll authoritatively quote the story: "All columns beside results_id (and, yes, this includes id) should have the type TEXT. This is because we must import all values provided, not just ones that match the spec's expectations."

danielglauser: Ahhh...yes. I did read that but thought that it wasn't necessary for boolean data. But if you want to report an error an include the original value then...`TEXT` it is.

danielglauser:
Instead of having composite primary keys would it make sense to make the `results_id` the primary key and put an index on the `id`? Not suggesting we change this here, rather just wondering if we've experimented with different options. Seems to be just a performance thing, not sure how important that is.

danielglauser:
How was this tested?

resources/migrations/20160509-create-5.0-csv-tables.up.sql
```diff
@@ -0,0 +1,285 @@
...
+                                      ballot_sub_title TEXT,
+                                      ballot_title TEXT,
+                                      electoral_district_id TEXT,
+                                      electorate_specification TEXT,
+                                      external_identifier_type TEXT,
+                                      external_identifier_othertype TEXT,
+                                      external_identifier_value TEXT,
+                                      has_rotation TEXT,
+                                      name TEXT,
+                                      sequence_order TEXT,
```
danielglauser: Given that there is a `sequence_order` on both the `candidate` and the `candidate_contests` is there really the same order, as in parallel arrays? If so should `candidate_contests` just have a foreign-key relationship with candidate or is this supposed to represent multiple contests per candidate?

tie-rack:
@danielglauser 

> Instead of having composite primary keys would it make sense to make the `results_id` the primary key and put an index on the `id`?

No. Primary keys must be unique, so just `results_id` is not enough.

Also, by making `(results_id, id)` the primary key, a  B-tree index is created. And since all our queries for data in these tables will be constrained by `results_id`, that index will act as an index on `id` as well.

resources/migrations/20160509-create-5.0-csv-tables.up.sql
tie-rack: `sequence_order` is being removed from `candidate` with the 5.1 changes (to come). This follows the 5.0 spec, which is what it should do.

danielglauser:
As long as you have tested this to the point you feel comfortable LGTM

sduckett:
@danielglauser 
> How was this tested?

I used `psql` to create, drop, and recreate the tables:

```sh
psql -d data_processor < resources/migrations/20160509-create-5.0-csv-tables.up.sql
psql -d data_processor < resources/migrations/20160509-create-5.0-csv-tables.down.sql
psql -d data_processor < resources/migrations/20160509-create-5.0-csv-tables.up.sql
```

PR #160, opened 2016-05-13
Create v5 data spec

sduckett:
Here are the data-specs that define the table names, columns, and any required coercions; the card mentions excluding coercions, but street segments already had a table that doesn't allow text values in some columns

[Pivotal card](https://www.pivotaltracker.com/story/show/118433873)

This is the second of three 5.0 CSV PRs, and blocks a PR about [loading 5.0 CSVs](https://github.com/votinginfoproject/data-processor/pull/161)


7maples:
LGTM - it was actually a lot easier to look at all of these together, fwiw

PR #162, opened 2016-05-17
Restart container on failure

tie-rack:
`--restart=always` wasn't working, and it interferes with the `Restart=on-failure` service directive.


sduckett:
LGTM, @tie-rack.

PR #163, opened 2016-05-19
Transform ballot measure contests

sduckett:
We've already loaded data from a set of CSV files, but it's kept in a table that looks a lot like a CSV. In order to generate some XML, we need to turn this flat, boring, data into a tree!

There's a set of functions in `vip.data-processor.db.translations.util` that forms the base of such translations; if these work correctly, then for each element in the XML spec, we can follow the approach used in `vip.data-processor.db.translations.v5-0.ballot-measure-contests` to define

- which rows to transform,
- the base path, and,
- how to build an entity

More details can be found in the [story @ Pivotal](https://www.pivotaltracker.com/story/show/118947361)


src/vip/data_processor/db/translations/util.clj
```diff
@@ -0,0 +1,113 @@
+(ns vip.data-processor.db.translations.util
+  (:require [clojure.string :as str]
+            [korma.core :as korma]
+            [vip.data-processor.db.postgres :as postgres]))
+
+(defn index-path [path]
+  (str path ".id"))
+
+(defn index-generator
```
tank157: This seems like the kind of thing a lazy sequence can do, but I'll trust your judgement that it needs to be a function.

tank157:
Ok, i can't claim to understand it all, but didn't see anything problematic, LGTM

sduckett:
@tank157 

> Ok, i can't claim to understand it all, but didn't see anything problematic, LGTM

Is there anything you're curious about?

tank157:
No, I'd have to delve too deeply into a domain I don't really need to know more about, it's cool.

src/vip/data_processor/db/translations/util.clj
sduckett: I'm trying to imagine what this would look like if it were written with lazy sequences; that didn't occur to me as an available approach.

tie-rack: For the trivial example in the docstring, a lazy sequence would be just fine. But we take this generator and pass it around to multiple functions that can take an arbitrary number of values from it (including zero). In order to use a lazy sequence instead, we would have to return the tail of the as-yet-used lazy sequence from those functions in addition to the results we're composing. That would result in, I believe, messier code.

tank157: Yeah, I didn't think it was truly viable, just noting that providing ever increasing numbers is often shown that way. This is better.

PR #165, opened 2016-05-23
Update schema from 5.0 to 5.1

sduckett:
We never supported v5.0, see? It's all v5.1 around here.

[Pivotal story](https://www.pivotaltracker.com/story/show/119862445)


tie-rack:
Looks like all the test-resources/xml/v5-*.xml files still say "5.0".

sduckett:
> Looks like all the test-resources/xml/v5-*.xml files still say "5.0".

I see that - it's not a quality we test.

tie-rack:
LGTM!

PR #166, opened 2016-05-24
Fix v5 full good run test

tie-rack:
There is an assertion that happens somewhere in the validating of xml files that the spec version of the loaded xml is in the known set of versions. We were not setting that in the test. Now we are.

Additionally, since this threw an exception that is caught by the pipeline, the test passed, despite not completing the pipeline. I've updated `assert-no-problems` to also check that the `:stop` and `:exception` keys on the context are nil. That would have caused the test to fail.


sduckett:
Thanks, LGTM.

PR #167, opened 2016-05-24
Update existing tables to match the 5.1 spec

sduckett:
The [5.1 XML spec](https://github.com/votinginfoproject/vip-specification/blob/vip511/vip_spec.xsd) has a set of changes described in a [Pivotal card](https://www.pivotaltracker.com/story/show/119862555).

This PR updates existing tables to match the spec where possible. Some tables are missing, some changes from the card aren't present - there are bugs filed to sort out this matter.

### DIFF
--- a/src/vip/data_processor/validation/data_spec/v5_1.clj
+++ b/src/vip/data_processor/validation/data_spec/v5_1.clj
@@ -76,17 +76,18 @@
    {:filename "ballot_selection.txt"
     :table :ballot-selections
     :columns [{:name "id"}
-              {:name "ballot_measure_contest_id"}
-              {:name "ballot_measure_contest_selection_id"}
+              {:name "ballot_measure_contest_ids"}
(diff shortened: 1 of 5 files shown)


src/vip/data_processor/validation/data_spec/v5_1.clj
```diff
@@ -110,6 +111,7 @@
               {:name "abbreviation"}
               {:name "ballot_sub_title"}
               {:name "ballot_title"}
+              ;; {:name "ballot_selection_ids" ;; this is in a join-table named 'candidate-contest-ballot-selections'
```
tie-rack: That table is dropped in this PR. I think we should have this column here.

sduckett: This and a few other things are being addressed - coming soon, @tie-rack.

tie-rack:
⛵

LGTM

PR #168, opened 2016-05-24
ballot_measure_selections to xml_tree_values

tie-rack:
BallotMeasureSelection! Take 5.1 CSV data and transform it as if it were imported from XML.

Pivotal story: [118947297](https://www.pivotaltracker.com/story/show/118947297)


sduckett:
LGTM, @tie-rack.

PR #169, opened 2016-05-24
Transform ballot_selection.txt to xml_tree_values

tie-rack:
Read from the v5_1_ballot_selections table and produce rows in the xml_tree_values table.

Pivotal story: [118947159](https://www.pivotaltracker.com/story/show/118947159)


sduckett:
LGTM, @tie-rack.

PR #171, opened 2016-05-25
Transform electoral_districts to xml_tree_values

tie-rack:
Read from the v5_1_electoral_districts table and produce rows in the xml_tree_values table.

Pivotal story: [119479119](https://www.pivotaltracker.com/story/show/119479119)


sduckett:
LGTM

PR #172, opened 2016-05-25
Transform candidate.txt to xml_tree_values

tie-rack:
Read from the v5_1_candidates table and produce rows in the xml_tree_values table.

Pivotal story: [119478659](https://www.pivotaltracker.com/story/show/119478659)


sduckett:
LGTM

PR #176, opened 2016-05-26
v5_1_localities -> xml_tree_values

sduckett:
Locality, locality, locality! Here it is as `xml_tree_values`

[Pivotal story](https://www.pivotaltracker.com/n/projects/1106394)


tie-rack:
LGTM

PR #178, opened 2016-05-26
Transform parties

sduckett:
🎉 🎈 🎂 
...but these are political parties we're talking about, and they're now transformed.

[Pivotal story](https://www.pivotaltracker.com/story/show/119861385)


tie-rack:
LGTM

PR #179, opened 2016-05-27
transform v5_1_party_contests -> xml_tree_values

sduckett:
Much like the other PRs, but this one features another table I missed earlier. (It seemed easier to fix it than file a bug.)

[Pivotal card](https://www.pivotaltracker.com/story/show/119861487)


tie-rack:
LGTM

PR #181, opened 2016-05-31
v5_1_polling_locations -> xml_tree_values

sduckett:
Here's a transformation of polling location to `xml_tree_values`. There's a new complex type for the `LatLng` element, which is similar to the `ExternalIdentifiers`, but with less nesting. [Pivotal card](https://www.pivotaltracker.com/story/show/119861935)

### DIFF
--- a/src/vip/data_processor/db/translations/util.clj
+++ b/src/vip/data_processor/db/translations/util.clj
@@ -94,8 +94,7 @@
 
 (defn latlng->ltree [idx-fn parent-path row]
   (when-not (and (str/blank? (:latitude row))
-                 (str/blank? (:longitude row))
-                 (str/blank? (:latlng_source row)))
+                 (str/blank? (:longitude row)))
     (let [index (idx-fn)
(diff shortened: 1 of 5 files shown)


src/vip/data_processor/db/translations/util.clj
```diff
@@ -92,6 +92,19 @@
                (simple-value->ltree :external_identifier_othertype "OtherType" parent-with-id)
                (simple-value->ltree :external_identifier_value "Value" parent-with-id)]))))
 
+(defn latlng->ltree [idx-fn parent-path row]
+  (when-not (and (str/blank? (:latitude row))
+                 (str/blank? (:longitude row))
+                 (str/blank? (:latlng_source row)))
```
tie-rack: Source is optional, so it shouldn't be required here.

sduckett:
An optional source - good catch, @tie-rack

tie-rack:
LGTM

PR #182, opened 2016-05-31
v5_1_precincts -> xml_tree_values

sduckett:
Some precincts can be transformed. [Pivotal card](https://www.pivotaltracker.com/story/show/119862047)


tie-rack:
LGTM

PR #188, opened 2016-06-02
Transform street segments

tie-rack:
Transforming street_segment.txt!

Since some street segment values are coerced to non-text values in its table, `simple-value->ltree` needed a tiny update to stringify values it deals with.

And `includes_all_streets` needed to be added to the table.

Pivotal story: [119862431](https://www.pivotaltracker.com/story/show/119862431)


sduckett:
LGTM

PR #189, opened 2016-06-03
Use Leiningen 2.5.3 -> 2.6.1 with Alpine Linux

sduckett:
Change the base image from a larger, [Debian base image](https://github.com/Quantisan/docker-clojure/blob/0d4a0abe13497a6081ebd080e83d1be0abab3f59/Dockerfile), to a smaller, [Alpine base image](https://github.com/Quantisan/docker-clojure/blob/0d4a0abe13497a6081ebd080e83d1be0abab3f59/alpine/Dockerfile). At the same time, upgrade Leiningen. You can see the [changes](https://github.com/technomancy/leiningen/blob/master/NEWS.md) between `2.5.3` and `2.6.1`.

I've tested this with out [data suite](https://github.com/votinginfoproject/data-suite) and the oft-used file `va-clean.zip`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM clojure:lein-2.5.3
+FROM clojure:lein-2.6.1-alpine
 
 RUN mkdir -p /usr/src/data-processor
 WORKDIR /usr/src/data-processor


cap10morgan:
I don't know @sduckett, I hear the maintainer of that upstream image is a real piece of work. Might want to think twice before using this.

LGTM

sduckett:
We can put it on staging to assuage your fears, @cap10morgan.

sduckett:
Do you have any thoughts on this, @tie-rack?

tie-rack:
Let's see this on staging!

PR #191, opened 2016-06-06
Fix a contact_information table-alias

sduckett:
These columns are aliased with a prefix of `:ci_` when the tables are
joined. I think we missed one, @tie-rack .


tie-rack:
Good catch!

tie-rack:
LG!

PR #192, opened 2016-06-07
Transform schedule.txt to HoursOpen(!) xml-tree-values

tie-rack:
HoursOpen is essentially just a container of Schedules, so it's possible to support them without hours_open.txt file. We can use schedule.txt and group by their `hours_open_id`....

Pivotal story: [119859211](https://www.pivotaltracker.com/story/show/119859211)


sduckett:
LGTM

PR #193, opened 2016-06-09
Transform election administration

sduckett:
In a sweeping finale, we have a _slightly more complex than usual_ ElectionAdministration element built from `department`, `contact_information`, and `voter_services` CSV files. This element has an unbounded number of departments, each of which has contact information and an unbounded number of voter services; it's safe to say that this could be a pretty broad tree.

[Pivotal card](https://www.pivotaltracker.com/story/show/119851529)


tie-rack:
Big LGTM!

PR #194, opened 2016-06-10
Add csv transform to pipeline

sduckett:
Happy Friday PR! 🎉 🎈 

The CSV->XML transformations we've previously written are now included in the processing pipeline. If you have a set of 5.1 CSV files to hand (there's a sample in `test-resources`), you can now load and transform them into `ltree` elements. This PR also makes a few updates to the way the pipeline is constructed, updates _all_ the tests we previously thought were good, and moves a few utilities from one namespace to another.

[Pivotal story](https://www.pivotaltracker.com/story/show/120374833)

Also - this branch is currently deployed to the VIP staging environment, so you could try sending a message via SQS and it would process them. Since there's no XML file output, it will fail with an exception when it tries to upload to S3, but that's a forthcoming story....


tie-rack:
At the risk of jumping the gun before it's assigned to me, this LGTM. 🍁

PR #195, opened 2016-06-13
Misc 5.1 fixes

tie-rack:
* People's Title and Profession weren't being transformed as InternationalizedText.
* Finish migrating PartySelection's PartyIds


sduckett:
LGTM, @tie-rack

PR #196, opened 2016-06-14
Validate that `xs:IDREFS` elements have referents

sduckett:
Some of the `xs:IDREF` elements changed to `xs:IDREFS` elements in the 5.1 spec, but not all of them. What was once a single element now is a space-delimited string. Links to specs, etc, are in the [Pivotal card](https://www.pivotaltracker.com/story/show/120708149)

This updates the validations to work with both types; the original query seemed useful, so I tried to find a way to turn a single row containing multiple IDs into multiple rows so I could use the same query.  I _think_ this is a pretty idiomatic way to use Postgres, but if not, let's discuss!

### DIFF
--- a/src/vip/data_processor/validation/v5.clj
+++ b/src/vip/data_processor/validation/v5.clj
@@ -33,7 +33,8 @@
    external-identifiers/validate-no-missing-values
    id/validate-unique-ids
    id/validate-no-missing-ids
-   id/validate-idrefs-refer
+   id/validate-idref-references
+   id/validate-idrefs-references
    precinct/validate-no-missing-names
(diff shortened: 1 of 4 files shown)


sduckett:
@ddellacosta knows some PostgreSQL, so I asked him to look at it with you, @tie-rack

ddellacosta:
Sean and I took a look, and I think that aside from any obvious performance problems there is a bit too much of a learning curve for me to comment in a timely fashion--for now I'm saying LGTM, and if we discover performance issues let's come back and review it again!

sduckett:
What about you, @tie-rack? Got any thoughts?

tie-rack:
LGTM!

PR #197, opened 2016-06-14
Add validations to 5.1 csv pipeline

tie-rack:
After a 5.1 CSV feed is loaded and transformed into xml_tree_values, use the 5.1 validations functions to validate the data!

`v5_1_ballot_styles` had yet to be transformed, so that is done, and one of its columns needed updating.

A "full good" CSV set is loaded and affirmed to be without errors. The data in that CSV set doesn't make a whole lot of sense, but it _is_ self-consistent and adheres to the relationships and value formats the spec demands.

Pivotal story: [120707911](https://www.pivotaltracker.com/story/show/120707911)


src/vip/data_processor/validation/xml.clj
```diff
@@ -244,7 +244,9 @@
     (with-open [reader (util/bom-safe-reader xml-file)]
       (let [vip-object (xml/parse reader)
             version (get-in vip-object [:attrs :schemaVersion])]
-        (assoc ctx :spec-version version)))))
```
sduckett: Oh yeah, we did this in `csv/determine-spec-version`...would that change have halted validations on any XML processed since that was deployed?

tie-rack: ...oh that is possible!

src/vip/data_processor.clj
```diff
@@ -34,16 +36,24 @@
...
+(defn add-validations
+  [{:keys [spec-version] :as ctx}]
+  (let [validations (condp = spec-version
+                      "3.0" v3-validation-pipeline
+                      "5.1" v5-1-validation-pipeline
+                      nil)]
+    (update ctx :pipeline (partial concat validations))))
+
 (def pipeline
   (concat download-pipeline
```
sduckett: I don't see `add-validations` called in this pipeline (but it's in the pipeline of `dev.core`); did I miss something?

tie-rack: Oh you're right. `add-validations` should replace the removed anonymous fn removed below. Commit coming.

sduckett: Ah, I see it now...`xml/version-pipelines` adds validations.

tie-rack: It shouldn't now, though.

sduckett:
OK, I think my questions and concerns are addressed. LGTM.

PR #198, opened 2016-06-20
Write 5.1 XML from CSV imports

tie-rack:
Now that 5.1 imports are translated into a tree that represents the resulting XML document, we can actually write that XML to a file.

The approach is essentially:

1. Pull the `xml_tree_values` rows out of the database in insert order.
2. Read through them, one by one, writing a fragment of XML for each row, closing and opening tags as necessary between rows.

By inserting the `xml_tree_values` rows in the order that the `write-xml` function needs them, we're able to avoid costly and complex ordering logic. A migration was added to include an auto-incrementing column (`SERIAL` Postgres type) which we `ORDER BY` when pulling the rows. To that end, rows for InternationalizedText elements needed their order swapped (attribute first, then data). Otherwise, we were already inserting in order!

The `write-xml` function is pretty stateful and procedural... but it's reading from a database and writing to a file. That's going to be stateful and procedural. Fortunately, it's not too long.

Pivotal stories: [120708367](https://www.pivotaltracker.com/story/show/120708367) and [121595323](https://www.pivotaltracker.com/story/show/121595323)

### DIFF
--- a/resources/migrations/20160616-add-insert-counter-to-xml-tree-values.down.sql
+++ b/resources/migrations/20160616-add-insert-counter-to-xml-tree-values.down.sql
@@ -1,0 +1,2 @@
+ALTER TABLE xml_tree_values
+DROP COLUMN insert_counter;
--- a/resources/migrations/20160616-add-insert-counter-to-xml-tree-values.up.sql
+++ b/resources/migrations/20160616-add-insert-counter-to-xml-tree-values.up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE xml_tree_values
+ADD COLUMN insert_counter BIGSERIAL;
--- a/src/vip/data_processor/db/postgres.clj
+++ b/src/vip/data_processor/db/postgres.clj
@@ -312,7 +312,7 @@
               (bulk-import ctx
                            (ent v3-0-import-entities) ; TODO: choose import-entities based on import version
                            (->> table
-                                (db.util/select-*-lazily 5000)
+                                (db.util/lazy-select 5000)
                                 (map #(assoc % :results_id import-id))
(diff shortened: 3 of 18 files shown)


cap10morgan:
Is there no library we can offload some of the nuts and bolts of the XML generation to (thinking of `clojure.data.xml/sexps-as-fragment` which is kind of Hiccup for XML)? All of the manual `<`, `>`, `=`, string concating, etc. leaves me feeling sad about maintaining this.

I can't claim to fully grok what the XML generation code is doing and why at every line, but it is tested, those tests pass, and I see no glaring issues besides my pondering above.

Assuming there isn't something that would be a slam-dunk improvement there, this LGTM.

tie-rack:
> Is there no library we can offload some of the nuts and bolts of the XML generation to (thinking of clojure.data.xml/sexps-as-fragment which is kind of Hiccup for XML)?

One nice thing about this approach versus building an intermediate data structure which is converted to XML is that we can stream directly from the database to the XML file with a constant memory overhead.

I'm afraid the complexity of building a series of fragments would be more complex than the 48-line `write-xml` function. That's the feeling I got when attempting it. Fortunately, our test case that works from database rows to generated XML covers every transition state between element attributes and element values, so future work has coverage.

And I'd like to note that Clojure's xml library has already failed us in another namespace and we had to write our own `emit-element` function anyway.

cap10morgan:
OK, that all makes sense. Unqualified LGTM, then.

sduckett:
I've got a few extras coming, @cap10morgan: when I tested against the "full good run", the upload to S3 failed.

cap10morgan:
Still LGTM

PR #200, opened 2016-06-27
UPPERCASE or lowercase headers; both work fine

sduckett:
While it might seem like a fun thing to nitpick, there's not semantic distinction based on the case-ness of a CSV file's headers. So, we ignore it by making everything lower-case and a little quieter. 📵 

[Pivotal story](https://www.pivotaltracker.com/story/show/117533533)


tie-rack:
Yep. This does the trick.

PR #202, opened 2016-06-28
Ignore duplicate electoral districts

tie-rack:
Pivotal story: [117504475](https://www.pivotaltracker.com/story/show/117504475)


sduckett:
LGTM

PR #203, opened 2016-06-28
XML validation errors are not fatal

sduckett:
Google is accepting XML that doesn't validate against the schema, so it
doesn't make sense to call it `:fatal`; let's call it an `:error`
instead.

[Pivotal Story](https://www.pivotaltracker.com/story/show/117504193)


sduckett:
...and it looks like this should be in `:errors` instead...

tie-rack:
There it is. 🎲 🎲

PR #204, opened 2016-07-05
Fix v5 public ids

sduckett:
`ElectionType` is an internationalized text element, so we need to adjust the ltree path. This lets us build a `public_id` that will be what Metis expects. Related to https://github.com/votinginfoproject/Metis/pull/320


tie-rack:
LGTM!

PR #206, opened 2016-08-16
Use bulk-import for an insert that may get v large

tie-rack:
An actual feed was blowing past the limit for an insert to Postgres. Using `bulk-import` keeps Postgres happy.


sduckett:
LGTM, @tie-rack 
🀄

PR #207, opened 2016-08-17
Handle csv errors in 5.1 feeds

tie-rack:
CSV errors were trying to put the line number as the "path" for errors in 5.1 feeds. Postgres does not like numbers in the place of an ltree path. So let's create a path for these kinds of errors.

Additionally, fix a lingering bug in the statistics query which was still trying to insert values into non-existent columns if the "element" didn't match a "value".

### DIFF
--- a/src/vip/data_processor/validation/csv.clj
+++ b/src/vip/data_processor/validation/csv.clj
@@ -105,8 +105,15 @@
                                   (let [message (.getMessage e)]
                                     (if (re-find #"UNIQUE constraint failed: (\w+).id" message)
                                       (db.util/retry-chunk-without-dupe-ids ctx sql-table chunk-values)
-                                      (assoc-in ctx [:fatal (:name sql-table) @line-number :unknown-sql-error]
-                                                [message]))))))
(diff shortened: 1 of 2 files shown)


src/vip/data_processor/validation/csv.clj
```diff
@@ -105,8 +105,15 @@
                                   (let [message (.getMessage e)]
                                     (if (re-find #"UNIQUE constraint failed: (\w+).id" message)
                                       (db.util/retry-chunk-without-dupe-ids ctx sql-table chunk-values)
-                                      (assoc-in ctx [:fatal (:name sql-table) @line-number :unknown-sql-error]
-                                                [message]))))))
+                                      (let [identifier (if (= "3.0" (:spec-version ctx))
+                                                         @line-number
+                                                         (str "CSVError.0."
```
sduckett: I'd think extract this, as a function of `filename` and `@line-number`.

tie-rack: Coming right up.

sduckett:
LGTM :bowtie:

PR #208, opened 2016-08-19
Fix identifier for number of values validation

tie-rack:
5.1 validations demand a path-like string for the identifier, not a number.


sduckett:
LGTM 🚙

sduckett:
...but isn't this the same as https://github.com/votinginfoproject/data-processor/commit/3cc3e1c07c96acb77e1372d04b0d72dda863ae23?

tie-rack:
It's not the same... It uses the function in that commit.

PR #209, opened 2016-08-29
Count `ElectionAdministration` errors as 3 types

sduckett:
Previously, errors with `Department` or `VoterService` elements were
included in the count of `ElectionAdministration` errors. This promotes
those errors so they show up in overview counts.

Fixes [this bug about grouping of elements](https://www.pivotaltracker.com/story/show/129071869) as well as [this bug about counts](https://www.pivotaltracker.com/story/show/129459125), where e.g., 1 `Source` element would be inflated to 5.

This blocks [a Metis PR](https://github.com/votinginfoproject/Metis/pull/330) that uses the database function.

### DIFF
--- a/resources/migrations/20160829-create-element-type-function.down.sql
+++ b/resources/migrations/20160829-create-element-type-function.down.sql
@@ -1,0 +1,1 @@
+DROP FUNCTION IF EXISTS element_type(path ltree);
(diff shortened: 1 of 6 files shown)


tie-rack:
LGTM 🛀

PR #210, opened 2016-08-30
ContactInformation is optional for VoterService

sduckett:
For a VoterService without any contact information, the final XML
would incorrectly omit it, but not any more!

[Bug](https://www.pivotaltracker.com/story/show/129489873)


tie-rack:
🎸

sduckett:
(Github thinks this is unmerged because I did it by hand; see https://github.com/votinginfoproject/data-processor/pull/209)

PR #211, opened 2016-09-06
Add `tablefunc` for use in dashboard queries

sduckett:
**Add `tablefunc` for use in dashboard queries**

This PR lets us use `crosstab` in some of the `ltree` based queries on the
dashboard.


tie-rack:
LGTM

PR #212, opened 2016-09-07
Create indices on (results_id, severity)

sduckett:
This is what fixed yesterday's slow query (& gateway timeout) experience.
Since these were already applied in production, they'll need to be dropped; I'll wait to merge until after hours.

Anything else you'd like to include, @tie-rack?


tie-rack:
You could add those `DROP INDEX IF EXISTS` to the up migration to avoid needing an after hours deploy.

sduckett:
> You could add those DROP INDEX IF EXISTS to the up migration to avoid needing an after hours deploy.

Right - it'll take around 40 minutes to create an index, so after hours is probably best.

tie-rack:
LGTM

PR #213, opened 2016-09-07
Fix polling_location.txt dependency check

tie-rack:
Instead of checking for precinct_polling_location.txt twice, also allow precinct_split_polling_location.txt to satisfy polling_location.txt's needs.


sduckett:
LGTM

PR #215, opened 2016-09-19
Index all the (seq scan) things

sduckett:
These simple indexes should make the dashboard's slowest-executing queries run much more efficiently. These are all related to that [sneaky bug where the dashboard hits a timeout](https://www.pivotaltracker.com/story/show/130475205).

### DIFF
--- a/resources/migrations/20160919-dashboard-performance-indexes.down.sql
+++ b/resources/migrations/20160919-dashboard-performance-indexes.down.sql
@@ -1,4 +1,4 @@
-drop index if exists street_segments_results_precinct_idx;
-drop index if exists precinct_polling_locations_results_precinct_idx;
-drop index if exists precinct_polling_locations_results_polling_location_idx;
+drop index if exists v3_0_street_segments_results_precinct_idx;
+drop index if exists v3_0_precinct_polling_locations_results_precinct_idx;
+drop index if exists v3_0_precinct_polling_locations_results_polling_location_idx;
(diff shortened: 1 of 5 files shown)


resources/migrations/20160919-dashboard-performance-indexes.up.sql
```diff
@@ -0,0 +1,11 @@
+drop index if exists v3_0_street_segments_results_precinct_idx;
+create index v3_0_street_segments_results_precinct_idx on v3_0_street_segments(results_id, precinct_id);
```
sduckett: This index came from street segment count queries being slow, and they don't use precinct splits. It's _likely_ useful somewhere, but it hasn't been measured.

resources/migrations/20160919-dashboard-performance-indexes.down.sql
```diff
@@ -0,0 +1,4 @@
+drop index if exists street_segments_results_precinct_idx;
+drop index if exists precinct_polling_locations_results_precinct_idx;
+drop index if exists precinct_polling_locations_results_polling_location_idx;
+drop index if exists v3_0_precincts_results_locality_idx;
```
sduckett: fixed.

resources/migrations/20160919-dashboard-performance-indexes.up.sql
tie-rack: Gotcha.

PR #218, opened 2016-09-26
Enable skipping feed validations

sduckett:
If the message sent over SQS contains `:skip-validations true`,
version-specific validation pipelines will be omitted from the pipeline.
While this might be useful in certain situations, the default is to keep
validations as part of the pipeline.


sduckett:
merged.

PR #220, opened 2016-09-29
Collect and process validation errors async

tie-rack:
# Async errors!

When a validation error is reported, instead of storing it on the context—which balloons and balloons as time goes on—instead, put it on a core.async channel. That channel is taken from and, every so often (once a number of messages is received or a timeout reached), the validation errors are committed to the database.

Because some errors are gathered before the specification version is known (and thus which table the errors must go to), the errors aren't processed until the spec version is decided. This was  accomplished by turning the spec-version into an atom with a watch.

This is a big one, and a long time coming. This may result in a very slight performance boost, but will definitely alleviate some OutOfMemory issues with particularly bad feeds.

### DIFF
--- a/dev-src/dev/core.clj
+++ b/dev-src/dev/core.clj
@@ -1,7 +1,11 @@
 (ns dev.core
   (:require [clojure.java.io :as io]
+            [clojure.core.async :as a]
+            [clojure.tools.logging :as log]
             [clojure.pprint :as pprint]
+            [utility-works.async :as util-async]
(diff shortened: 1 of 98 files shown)


sduckett:
This depends on https://github.com/democracyworks/utility-works/pull/1, right?

tie-rack:
@sduckett Yes. That PR's code is currently out as 0.1.0-SNAPSHOT on Clojars.

PR #221, opened 2016-09-29
Break no-missing validator down into 2 queries

cap10morgan:
Each is far simpler & faster. One gets expected, the other actual, then
we just run `clojure.set/difference` on them.

### DIFF
--- a/src/vip/data_processor/validation/v5/util.clj
+++ b/src/vip/data_processor/validation/v5/util.clj
@@ -3,7 +3,8 @@
             [vip.data-processor.db.postgres :as postgres]
             [vip.data-processor.validation.xml.spec :as spec]
             [clojure.string :as str]
-            [clojure.tools.logging :as log]))
+            [clojure.tools.logging :as log]
+            [clojure.set :as set]))
 
 (defn two-import-ids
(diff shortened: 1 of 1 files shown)


src/vip/data_processor/validation/v5/util.clj
```diff
@@ -69,27 +70,55 @@
...
+                              ["SELECT DISTINCT subltree(path, 0, CAST(? AS INT)) AS path
+                                FROM xml_tree_values
+                                WHERE results_id = ?
+                                AND simple_path <@ text2ltree(?)"
+                               [path-to-parent-nlevel import-id
+                                path-to-parent]]
+                              :results))
+                children (map (fn [r]
+                                {:parent (-> r :parent_path .getValue)
+                                 :child (-> r :child_path .getValue)})
```
sduckett: It appears that `:child` is not used anywhere else; after building `found-parent-set` on line 106, it's just a set of `:parent_path`s, right?

cap10morgan: Yes that's right. I can drop child and simplify this.

sduckett:
LGTM, so let's see how it runs! 👟

PR #227, opened 2016-10-07
Include the state's name as part of the filename

sduckett:
The idea is that this should help catch state / fips code mismatches. Below, you see what ends up in the processed bucket when a Virginia feed decides to impersonate Florida (FIPS 12).

screen shot 2016-10-07 at 11 25 56 am

### DIFF
--- a/src/vip/data_processor/s3.clj
+++ b/src/vip/data_processor/s3.clj
@@ -44,10 +44,17 @@
     (< (count fips) 5) (format "%05d" (Integer/parseInt fips))
     :else fips))
 
-(defn zip-filename* [fips election-date]
+(defn format-state
+  [state]
+  (if (nil? state)
+    "YY"
+    (clojure.string/replace (clojure.string/trim state) #"\s" "-")))
+
+(defn zip-filename* [fips state election-date]
   (let [fips (format-fips fips)
+        state (format-state state)
         date (util/format-date election-date)]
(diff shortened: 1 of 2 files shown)


src/vip/data_processor/s3.clj
```diff
@@ -44,10 +44,17 @@
...
+  (if (nil? state)
+    "YY"
+    (clojure.string/replace (clojure.string/trim state) #"\s" "-")))
+
+(defn zip-filename* [fips state election-date]
   (let [fips (format-fips fips)
+        state (format-state state)
         date (util/format-date election-date)]
-    (str (join "-" ["vipfeed" fips date]) ".zip")))
+    (str (join "-" ["vipfeed" fips state date]) ".zip")))
```
tie-rack: @jungshadow This is only for the name of the zip file. The name of the XML is unchanged.

For county-based feeds, the fips used _should_ be the 5-digit county fips.

jungshadow: > This is only for the name of the zip file. The name of the XML is unchanged.

@tie-rack Sorry if I wasn't clear. The zip file naming convention is also covered in the standard (see the linked section).

> For county-based feeds, the fips used should be the 5-digit county fips.

To provide a quick example for my edification, the zip filename for a Maricopa County, AZ feed for the general election would be ```vipFeed-04013-AZ-2016-11-08.zip```? That looks fine to me.

PR #229, opened 2016-10-10
Support 5.1.1 and 5.1.2 imports

tie-rack:
Allow importing of 5.1.1 and 5.1.2 feeds, validating against the spec determined by the import.

In many places, however, the patch level does not matter, so ignore it there.


sduckett:
in `postgres.clj`:
```
 (defn store-spec-version [{:keys [spec-version import-id] :as ctx}]
   (when @spec-version
     (korma/update results
       (korma/set-fields {:spec_version @spec-version})
       (korma/where {:id import-id})))
```

Should we be storing the patch level here? I can see where it could be useful as a debugging aid, but also where it could be an unnecessary distinction.

tie-rack:
@sduckett To the first question: yes, we should be able to remove the 5.0 spec, since it's not used anywhere. I'll do that.

To the second question: The spec version with the patch level is used to determine the XML spec that drives some validations. I think we should keep the patch level since it has a real effect on validations.

sduckett:
> To the second question: The spec version with the patch level is used to determine the XML spec that drives some validations. I think we should keep the patch level since it has a real effect on validations.

For validations, yes, we'd need the patch version, but we use whatever is contained in the `spec-version` atom; we never return to the `results` table to drive our validation choices.

PR #232, opened 2016-10-11
Safer version without patch

sduckett:
Sometimes people make mistakes and a feed will be missing the spec-version. If that happens, we should stop processing because "no version" is an "unsupported version" instead of throwing an exception.

### DIFF
--- a/test/vip/data_processor/util_test.clj
+++ b/test/vip/data_processor/util_test.clj
@@ -35,3 +35,17 @@
   (testing "when we run into a vector, that becomes part of the value"
     (is (= {[:to :flatten :or] [:not {:to "flatten"}]}
            (flatten-keys {:to {:flatten {:or [:not {:to "flatten"}]}}})))))
+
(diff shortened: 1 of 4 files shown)


test/vip/data_processor/util_test.clj
```diff
@@ -35,3 +35,17 @@
...
+    (is (nil? (version-without-patch ""))))
+
+  (testing "with a two-part version number, we get our input back"
+    (is (= "5.1" (version-without-patch "5.1"))))
+
+  (testing "with a three-part version number, we strip the patch level"
+    (is (= "5.1" (version-without-patch "5.1.1"))))
+
+  (testing "going past three levels might surprise you"
+    (is (= "5.1.2.3" (version-without-patch "5.1.2.3")))))
```
tie-rack: I would rather call this unspecified behavior than add a test that asserts this behavior.

sduckett: Deal.

PR #234, opened 2016-10-24
Individual locality views

sduckett:
In this [Pivotal card](https://www.pivotaltracker.com/story/show/123274961), the dashboard gets a view into a 5.1 feeds by locality. This PR contains the necessary database pieces to make that happen _reasonably_ quickly. For small localities (e.g., those containing < 50 precincts), it is fast enough, but as the numbers of precincts (and simultaneously, polling locations and street segments) increases, the query time also increases. For Wake County, NC, which contains 495 precincts, it takes about 10 seconds. This might be the sort of thing we continue to optimize or perhaps we summarize by locality as we process the feed, but I'm more interested in correctness at this stage.

The function `element_type` is now `IMMUTABLE`. This allows it to be used as part of an index; functions are presumed to be `VOLATILE` by default. This was the largest contributor to getting the whole `locality_stats` function on the order of seconds rather than minutes.

### DIFF
--- a/resources/migrations/20161011-01-add-completion-rate-function.down.sql
+++ b/resources/migrations/20161011-01-add-completion-rate-function.down.sql
@@ -1,0 +1,1 @@
+drop function completion_rate(row_count bigint, error_count bigint);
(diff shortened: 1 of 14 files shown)


sduckett:
There's a better way - [there's a new PR](https://github.com/votinginfoproject/data-processor/pull/243) that extends / replaces this one.

PR #242, opened 2016-12-05
Validate the format of dates

sduckett:
The `Schedule` elements of a feed should have dates that are in the form `YYYY-MM-DD`; this PR adds a validation that checks their format.

[Pivotal card](https://www.pivotaltracker.com/story/show/128890921)


test/vip/data_processor/validation/v5/hours_open_test.clj
```diff
@@ -8,6 +8,40 @@
...
+                          {:severity :errors
+                           :scope :hours-open
+                           :identifier "VipObject.0.HoursOpen.0.Schedule.0.StartDate.1"})
+
+      (assert-no-problems errors
+                          {:severity :errors
+                           :scope :hours-open
+                           :identifier "VipObject.0.HoursOpen.0.Schedule.0.EndDate.2"}))
+
+    (testing "a start date or end date that looks like a timestamp is invalid"
```
cap10morgan: This appears to be an apt description of the first test in this `testing` block, but the second one is a semantically-invalid date, no?

sduckett: It looks like the W3 folks consider `2019-02-29` to be wrongly formatted since they tie semantics into their definition of `xs:date`: https://www.w3.org/TR/xmlschema-2/#isoformats, which means we'll have to go a little farther than a regex.

cap10morgan:
`clj-time` to the rescue!

PR #243, opened 2016-12-16
Summarize locality details for a feed

sduckett:
This [Pivotal story](https://www.pivotaltracker.com/story/show/123274961) asked for an overview of a feed by locality; an [older PR](https://github.com/votinginfoproject/data-processor/pull/234) from before election day built the base - and was approved - but it wasn't very performant. 

This PR changes that. Instead of a function that gets stats for just _one_ locality, we break the feed down by localities as part of the processing. Most of the changes are renaming migrations to prevent conflicts (we got a few in between the original PR and this one).


resources/migrations/20161206-04-create-locality-stats-fn.up.sql
```diff
@@ -0,0 +1,197 @@
...
+
+    polling_location_count := cardinality(precinct_polling_locations);
+    polling_location_completion := v5_dashboard.completion_rate(polling_location_count, polling_location_errors);
+
+    select count(subpath(errors.path, 0, 4))
+    into street_segment_errors
+    from xml_tree_validations errors
+    where errors.results_id = rid
+      and errors.path <@ street_segments
+      and errors.path ~ '*.StreetSegment.*';
```
tie-rack: These kinds of ltree searches make me a little nervous. How do queries with clauses like `~ '*.something.*'` perform in a well-populated db?

katrinamariehh:
Is there or will there be a corresponding story on Metis so we can see this data on the dashboard?

sduckett:
> Is there or will there be a corresponding story on Metis so we can see this data on the dashboard?

Yes, @katrinamariehh, there is [a corresponding story](https://github.com/votinginfoproject/Metis/pull/340).

tank157:
It's difficult for me to understand what's going on in all of this, it's almost entirely migrations. Are there tests elsewhere?

sduckett:
> It's difficult for me to understand what's going on in all of this, it's almost entirely migrations. Are there tests elsewhere?

Good point; I bet I can cook some up, @tank157.

PR #245, opened 2016-12-21
house number validation

katrinamariehh:
Added validation for checking that house numbers are positive integers for this [Pivotal story](https://www.pivotaltracker.com/story/show/126604869). Validation function is in value_format.clj; it uses the all-digits check first, then coerces the string to an integer and confirms that the integer is greater than or equal to zero. It's tested in value_format_test.clj (I included a test for all-digits that I wrote to understand testing and since there wasn't already one I left it).

### DIFF
--- a/src/vip/data_processor/validation/data_spec/value_format.clj
+++ b/src/vip/data_processor/validation/data_spec/value_format.clj
@@ -1,5 +1,4 @@
-(ns vip.data-processor.validation.data-spec.value-format
-  (:require [vip.data-processor.validation.data-spec.coerce :as coerce]))
+(ns vip.data-processor.validation.data-spec.value-format)
 
 (def all-digits
   {:check #"\A\d+\z"
--- a/src/vip/data_processor/validation/data_spec/value_format.clj
+++ b/src/vip/data_processor/validation/data_spec/value_format.clj
@@ -6,17 +6,6 @@
    :message "Invalid data type"
    :severity :fatal})
 
-(defn check-not-negative [val]
-  (let [house-number (coerce/coerce-integer val)]
-      (and
-       (re-find (:check all-digits) val)
-       (>= house-number 0))))
-
-(def not-negative-integer
-  {:check check-not-negative
-   :message "House numbers must consist of postive integers only"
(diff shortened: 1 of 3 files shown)


src/vip/data_processor/validation/data_spec/value_format.clj
```diff
@@ -1,10 +1,22 @@
...
 (def all-digits
   {:check #"\A\d+\z"
    :message "Invalid data type"
    :severity :fatal})
 
+(defn check-not-negative [val]
+  (let [house-number (coerce/coerce-integer val)]
+      (and
+       (re-find (:check all-digits) val)
+       (>= house-number 0))))
```
tie-rack: If `coerce-integer` returns `nil`, which it can, then this line with throw an exception because you can't compare `nil` to 0.

katrinamariehh: To answer you first question--that's totally what I thought at first, too!  And then @sduckett was all like, "But what about NEGATIVE NUMBERS."  So this is specifically to make sure all the house numbers are positive integers.  Your second question about nil, if `(re-find (:check all-digits) val)` is `false` it doesn't evaluate the second part of the `and` and just returns `false`.

tie-rack: You're right, it'll short-circuit out first... but there's still no way to have the `all-digits` check succeed and end up with a negative number for the `>=` to fail on. It's redundant.

katrinamariehh: Now that I look at the test I think you're right; I thought I put test data through and it did not fail, but now I'm just not sure.  I'll take a look and get back to you!

katrinamariehh: You're totally right! I *thought* I had checked before starting about negative integers, but clearly I did not!

katrinamariehh: I'm going to remove my changes but add tests to confirm the expected behavior (that negative house numbers will be caught).

PR #247, opened 2017-01-03
The master branch has the most up to date schema

sduckett:
It seems easier to fix the reference than keep a handful of branches synchronized.
[pivotal story](https://www.pivotaltracker.com/story/show/131325151)

### DIFF
--- a/src/vip/data_processor/output/tree_xml.clj
+++ b/src/vip/data_processor/output/tree_xml.clj
@@ -92,7 +92,7 @@
 
 (defn write-xml [file spec-version import-id]
   (with-open [f (io/writer (.toFile file))]
-    (.write f (str "<?xml version=\"1.0\"?>\n<VipObject xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" schemaVersion=\"" spec-version "\" xsi:noNamespaceSchemaLocation=\"http://votinginfoproject.github.com/vip-specification/vip_spec.xsd\">\n"))
+    (.write f (str "<?xml version=\"1.0\"?>\n<VipObject xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" schemaVersion=\"" spec-version "\" xsi:noNamespaceSchemaLocation=\"https://raw.githubusercontent.com/votinginfoproject/vip-specification/master/vip_spec.xsd\">\n"))
     (let [last-seen-path (atom "VipObject.0")
           inside-open-tag (atom false)
           values-written (atom 0)]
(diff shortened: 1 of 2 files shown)


src/vip/data_processor/output/tree_xml.clj
```diff
@@ -92,7 +92,7 @@
 
 (defn write-xml [file spec-version import-id]
   (with-open [f (io/writer (.toFile file))]
-    (.write f (str "<?xml version=\"1.0\"?>\n<VipObject xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" schemaVersion=\"" spec-version "\" xsi:noNamespaceSchemaLocation=\"http://votinginfoproject.github.com/vip-specification/vip_spec.xsd\">\n"))
+    (.write f (str "<?xml version=\"1.0\"?>\n<VipObject xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" schemaVersion=\"" spec-version "\" xsi:noNamespaceSchemaLocation=\"https://raw.githubusercontent.com/votinginfoproject/vip-specification/master/vip_spec.xsd\">\n"))
```
tie-rack: Should the `xsi:noNamespaceSchemaLocation` be set to the location of the spec based on the `spec-version`? if it's 3.0, say, the master branch's vip_spec.xsd won't be appropriate.

sduckett: > Should the xsi:noNamespaceSchemaLocation be set to the location of the spec based on the spec-version?

We are only writing 5.x here, but there are still some slight differences to consider.

sduckett:
📟  @tie-rack; what do you think about selecting a tag based on the `spec-version`?

tie-rack:
A tag isn't quite enough. I think you need to select the whole url. For 3.0, the file isn't vip_spec.xsd. For 5.1.2, the tagged version still has 5.1.1 as its version.

I'd suggest
* For 3.0, use https://github.com/votinginfoproject/vip-specification/blob/v3-archive/vip_spec_v3.0.xsd
* For 5.1.2, make a branch starting from the v5.1.2-release tag, push a commit that updates the version number to 5.1.2, tag _that_ as something, and use that tag's vip_spec.xsd.

sduckett:
This NS doesn't write 3.0 xml, @tie-rack, so I'm not sure the first point applies. For the second, a commit got into master a couple days ago and the tag needs to be adjusted to reflect it.

tie-rack:
Yep, forgot this was 5.x only. So, yeah, if the tag is adjusted (and there's no diff between the current release tag and the new one besides the updated version number) that'll work.

sduckett:
The tag `v5.1.2-release` was updated (https://github.com/votinginfoproject/vip-specification/issues/356); anything else you need @tie-rack?

PR #249, opened 2017-01-10
Build the public id with warm, safe, US-ASCII

sduckett:
We build a `public_id` value to use in our URLs, but it's coming from user supplied data, which sometimes might contain characters that break routing in the dashboard. It seemed like percent encoding the string would be the best solution, but
something in the dashboard isn't properly handling those encoded strings, so
the next best solution seems to be removing things that would cause it
to break.

[Story A](https://www.pivotaltracker.com/story/show/132265201), [story B](https://www.pivotaltracker.com/story/show/132272105)

### DIFF
--- a/src/vip/data_processor/db/postgres.clj
+++ b/src/vip/data_processor/db/postgres.clj
@@ -10,7 +10,8 @@
             [vip.data-processor.util :as util]
             [vip.data-processor.validation.data-spec :as data-spec]
             [clojure.set :as set])
-  (:import [org.postgresql.util PGobject]))
+  (:import [org.postgresql.util PGobject]
+           [java.text Normalizer]))
 
 (defn url []
   (let [{:keys [host port user password database]} (config [:postgres])]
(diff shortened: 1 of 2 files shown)


src/vip/data_processor/db/postgres.clj
```diff
@@ -134,7 +135,11 @@
         formatted-date (util/format-date date)
         good-parts (->> [formatted-date election-type state]
                         (remove nil-or-empty?)
-                        (map #(str/trim %)))]
+                        (map #(str/trim %))
+                        (map #(Normalizer/normalize % java.text.Normalizer$Form/NFKD))
+                        (map #(str/replace % #"\p{Space}" "-"))
+                        (map #(str/replace % #"[\p{Punct}&&[^-]]" "%"))
+                        (map #(str/replace % #"[\P{ASCII}%]" "")))]
```
tie-rack: Why convert punctuation (except `-`) to `%` and then in the next line convert `%` to nothing? Couldn't that just go straight to nothing without the intermediary `%`?

sduckett: The date string is written like `2017-01-11`, which feels pretty readable.

tie-rack: My question is about replacing some characters with `%` and then replacing `%` with nothing. Why not replace those characters with nothing in the first place?

sduckett: Oh, you're right; there is an unnecessary step there 🥇

PR #252, opened 2017-02-22
Don't force apartment numbers to be integers

sduckett:
In [this story](https://www.pivotaltracker.com/story/show/137898631), we thought it would be trivial to remove a validation on apartment numbers. However, when writing XML for a 3.0 feed, we validate it against the spec, so we'd need to change the XSD as well if strings are going to be allowed (ok, it's a little less trivial). Also, I noticed the pipeline used in `xml_test/write-xml-test` was letting errors sneak through with passing tests (see [this chore](https://www.pivotaltracker.com/story/show/135355555))

So, here is a little PR that introduces `3.0.1`, for your perusal. I'll make an issue / PR for [vip-specification](https://github.com/votinginfoproject/vip-specification), but for now, how does this approach feel? Is there a better way?

### DIFF
--- a/resources/specs/vip_spec_v3.0.xsd
+++ b/resources/specs/vip_spec_v3.0.xsd
@@ -268,8 +268,8 @@
               <xs:element name="start_house_number" type="xs:integer" />
               <xs:element name="end_house_number" type="xs:integer" />
               <xs:element name="odd_even_both" type="oebEnum" />
-              <xs:element minOccurs="0" name="start_apartment_number" type="xs:integer" />
-              <xs:element minOccurs="0" name="end_apartment_number" type="xs:integer" />
+              <xs:element minOccurs="0" name="start_apartment_number" type="xs:string" />
+              <xs:element minOccurs="0" name="end_apartment_number" type="xs:string" />
               <xs:element name="non_house_address" type="detailAddressType" />
               <xs:element name="precinct_id" type="xs:integer" />
               <xs:element minOccurs="0" name="precinct_split_id" type="xs:integer" />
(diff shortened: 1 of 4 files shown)


MariaBianchi:
I think that we can suggest the change to the spec, though I'm not sure how likely changes to the 3.0 spec are to be made at this point. I've wondered about Chris's suggestion (should we bother importing apartment #s at all, since Google isn't using them) in the past, and have always landed on the side of keeping them in because A.) At some point someone else might use the data for something and B.) If we drop apartment numbers, duplicates that are just weird duplicate rows and duplicates caused by apartments will look identical to us, which might make communications around error resolution with the state more confused. I'm not sure either of these reasons is convincing enough not to do it. @afsmythe, what do you think about this?

afsmythe:
I think I'd prefer to have the spec updated. @MariaBianchi makes a good point about what to do with duplicate street_segments once the apartment number is ignored. For Washington, during the general election, I was having to strip the apartment numbers from the feed and then de-duplicate the resulting street_segment records. Why? I don't recall, but maybe Google was ignoring the duplicates. Ignoring apartment numbers that are invalid (because of alpha characters) makes sense so long as the result is de-duplicated. But, alpha apartment identifiers are common and should be allowed by the spec, IMO.

sduckett:
It looks to be trivially easy to just ignore (`4a` -> `null`; `42` -> `42`) any of the apartment numbers that are "invalid" as @tie-rack suggested

@MariaBianchi makes a good point about changing an old, deprecated spec; I suppose we could change our implementation and be divergent if the apartment number idea doesn't go over, but that doesn't feel right.

sduckett:
> I wonder if instead of creating a new version of the spec (which doesn't track anything official), it would make more sense to not output apartment numbers if they're invalid

@tie-rack: After talking with @MariaBianchi, it sounds like it might be better to keep the apartment numbers around and call them strings. I also rearranged the commits to be a bit more clear.

PR #255, opened 2017-03-20
Apartment numbers are text

katrinamariehh:
Related to [this story](https://www.pivotaltracker.com/story/show/137898631)--we changed the spec to allow `start_apartment_number` and `end_apartment_number` but didn't change the `v3_0_street_segments` table to store those columns as text.  This PR includes an up migration that alters the table schema to have those columns be text and an slightly more complicated down migration that checks if it can cast those columns as integers before changing the columns back to being integers.

### DIFF
--- a/resources/migrations/20170316-00-v3-apartment-number-strings.down.sql
+++ b/resources/migrations/20170316-00-v3-apartment-number-strings.down.sql
@@ -1,2 +1,21 @@
-alter table v3_0_street_segments alter column start_apartment_number set data type integer using start_apartment_number::integer;
-alter table v3_0_street_segments alter column end_apartment_number set data type integer using end_apartment_number::integer;
+/*update v3_0_street_segments set start_apartment_number = null where start_apartment_number = '';*/
+/*update v3_0_street_segments set end_apartment_number = null where end_apartment_number = '';*/
+
+create or replace function convert_to_integer(apt_number_txt text)
+returns integer as $$
+declare apt_number_int integer default null;
+begin
+    begin
+        apt_number_int := apt_number_txt::integer;
+    exception when others then
+        return null;
+    end;
+return apt_number_int;
+end;
+$$ language plpgsql;
+
+
+alter table v3_0_street_segments alter column start_apartment_number set data type integer using convert_to_integer(start_apartment_number);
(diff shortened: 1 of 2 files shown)


resources/migrations/20170316-00-v3-apartment-number-strings.down.sql
```diff
@@ -0,0 +1,18 @@
+create or replace function convert_to_integer(apt_number_txt text)
+returns integer as $$
+declare apt_number_int integer default null;
+begin
```
sduckett: Lines 4-11 show an interesting characteristic of PostgreSQL: [`begin`](https://www.postgresql.org/docs/9.4/static/sql-begin.html) and [`end`](https://www.postgresql.org/docs/9.4/static/sql-end.html) have transaction semantics when you're writing *SQL* with Postgres. When you're writing *PL/pgSQL*, as when you create a new function, [begin/end form a block for grouping statements](https://www.postgresql.org/docs/9.4/static/plpgsql-structure.html) instead of a transaction. There's a sneaky implicit transaction going on, but it's at the function call level:

> Note: There is actually a hidden "outer block" surrounding the body of any PL/pgSQL function. This block provides the declarations of the function's parameters (if any), as well as some special variables such as FOUND (see Section 40.5.5). The outer block is labeled with the function's name, meaning that parameters and special variables can be qualified with the function's name.

The docs go on to say that begin/end mean two different things in different contexts:

> It is important not to confuse the use of BEGIN/END for grouping statements in PL/pgSQL with the similarly-named SQL commands for transaction control. PL/pgSQL's BEGIN/END are only for grouping; they do not start or end a transaction. Functions and trigger procedures are always executed within a transaction established by an outer query — they cannot start or commit that transaction, since there would be no context for them to execute in. However, a block containing an EXCEPTION clause effectively forms a subtransaction that can be rolled back without affecting the outer transaction. For more about that see [Section 40.6.6.](https://www.postgresql.org/docs/9.4/static/plpgsql-control-structures.html#PLPGSQL-ERROR-TRAPPING)

This means you could write one less begin/end pair and have the same semantics and behavior:

```sql
create or replace function convert_to_integer(apt_number_txt text)
returns integer as $$
declare apt_number_int integer default null;
begin
  apt_number_int := apt_number_txt::integer;
  exception when others then return null;
  return apt_number_int;
end;
$$ language plpgsql;
```

PR #256, opened 2017-03-22
Feeds by election, on cascade delete

katrinamariehh:
This is part of covering [this story](https://www.pivotaltracker.com/story/show/128890329) about deleting old feeds from the database.  This PR covers a structural change in the database to facilitate the process of deleting (by setting the `foreign key` constraints to `on cascade delete` we can remove all related data for a feed by deleting from the `results` table that all of the other tables have a relationship with) as well as a function to aggregate feeds by election where feeds are the "same" if they have the same `vip_id`, `name`, `election_type`, and `election_date`.  Per Franklin's request, `cleanup.feeds_by_election` chunks the 5 most recent version of a feed and then all the feeds older than that for deletion.

The process of actually deleting the old data is not included in this work and will be taken care of separately, these changes are just to make it possible/easy to delete old data.


resources/migrations/20170320-02-cleanup-feeds-by-election.down.sql
```diff
@@ -0,0 +1,2 @@
+drop schema cleanup;
```
sduckett: It might be a good idea to say `drop schema if exists cleanup`; between 9.4 and 9.6, `if exists` was added to a few places, but I forgot [it was already here](https://www.postgresql.org/docs/9.4/static/sql-dropschema.html)

resources/migrations/20170320-02-cleanup-feeds-by-election.up.sql
```diff
@@ -0,0 +1,39 @@
+create schema cleanup;
```
sduckett: As with [dropping the schema](https://github.com/votinginfoproject/data-processor/pull/256/files#diff-57449c656bca65c4da048ed094f78b79R1), we might say [`create schema if not exists cleanup`](https://www.postgresql.org/docs/9.4/static/sql-createschema.html)

tank157: If the schema already existed, would we want to run the create.view as is?

resources/migrations/20170320-02-cleanup-feeds-by-election.down.sql
katrinamariehh: makes sense!  I'll update it.

PR #258, opened 2017-03-27
Replacing materialized views with tables

katrinamariehh:
This is a PR for [this story](https://www.pivotaltracker.com/story/show/141620183) (it's a research card, but researching the approach resulted in the work being completed).  In order to cut down on processing time we've narrowed down that refreshing the materialized views is very time-consuming.  We decided to replace the views with tables and then replace the refresh step in the pipeline with a function to populate the tables for a given feed.  Since we are now running a query for just one feed rather than against all of the data in the `xml_tree_values` table, we should be able to decrease processing time in production a good bit (in some cases it looks like 1/3 of the processing time is spent refreshing the materialized views).

The query that runs to populate the `v5_dashboard.elections` table looks very similar to the materialized view `v5_dashboard.elections` but it is narrowed down to a single `results_id` on the `xml_tree_values` table rather than running against the entire table.  To confirm this behavior would result in faster processing, I ran 60 feeds on the master branch and then 60 feeds on this branch and compared the results.  By the end of the run the `refresh materialized views` step of the pipeline was taking 15 seconds to run on master and the replacement (running functions to populate the tables) took less than a second for all 60 feeds.


src/vip/data_processor/db/postgres.clj
```diff
@@ -275,11 +275,41 @@
      [(str "refresh materialized view " view)]))
   ctx)
 
+(defn populate-i18n-table
```
sduckett: One Clojure idiom that's pretty fun to use is [destructuring](https://clojure.org/guides/destructuring), which succinctly combines the function's arguments and the `let` form that tends to follow `map` arguments, like this:

```clojure
(defn populate-i18n-table
  [{:keys [import-id] :as ctx}]
  (log/info "Populating v5_dashboard.i18n table"
  (korma/exec-raw
    (:conn xml-tree-values)
    ["select v5_dashboard.populate_i18n_table(?)" [id]])))
  ctx)
```

katrinamariehh: Cool--I can definitely make this change and I see where it's a little neater than using a let in these cases.  I'm also thinking I should delete the `refresh-materialized-views` function since it's not longer something that can be called.

PR #259, opened 2017-03-30
Cursors are great for performance, but we should use them properly

sduckett:
Sometimes [our database tests deadlock][bug] when we run them on our dev machines, but they don't deadlock when running inside a Docker container with `script/postgres-tests`. This works, but it is a little slow - and marginally annoying - to have to run the tests in Docker all of the time. Taking a deeper look - and `tail`ing your `postgres.log` file - you'd notice that when running any tests around any code that use a database cursor - for performance reasons - the code would retry `fetch`es and `close`s; sometimes it can't close the cursor, so it gives up to avoid spinning indefinitely.

That doesn't seem like the right thing to do! I'd like to get rid of this deadlock so I don't _have to_ run tests only in a Docker container. Second, we should use cursors more idiomatically, since they exist per connection rather than globally (this fact is either buried in the Postgres manual, or pretty easy so demonstrate with two `psql` sessions side by side.)

### RIP Deadlock

Updating database related libraries resolved the deadlock; it appears to be fixed [somewhere between clojure.java.jdbc "0.3.7" and "0.6.1"][jdbc], but the cursor was still extant and fetchable only once in a while.

### Avoid `*print-length*`

If I recall correctly, while @tie-rack and I were working on CSV translations, some of the errors were extremely long and trashed our CIDER sessions. The fix at the time was to set `*print-length*` in `test-helpers`, but that causes an `IllegalStateException` when reloading altered namespaces with CIDER. By putting it in the project file, we don't have to restart CIDER all the time.

### Use a cursor with a single connection

I started by making the connection pool only hold one connection, which worked wonderfully, but at the cost of making future parallelization of anything related to databases impossible. Next, I tried to bind to a single connection inside `lazy-select-fn`, but time and again, I was fetching from the wrong connection.

Korma wasn't doing the right thing, so I reached for `clojure.java.jdbc/get-connection` hoping to leave the useful connection pool in place and _really, truly_ have one connection for working with the cursor. I fetched the first chunk and felt a sense of elation; on the second fetch, it came crashing down with a `java.sql.SQLException`, telling me the connection had already been closed. After some hand-wringing, I decided my understanding of laziness was upside down; we'd returned beautiful lazy sequence, left the scope where the connection was bound, and then fetched one chunk. After that fetch, there wouldn't have been any reference to the connection, so it could be closed. Could it be related to using `trampoline`? I swear I fetched many chunks when I was at a REPL...

The function `write-xml` consumes a lazy sequence of values to generate XML tags, so if we move the connection binding into `generate-xml`, we'll always have the connection we need. We're also using a cursor to zip through the very large (proportionally) set of street segments, so I
used the same trick to adjust its `transformer`.

The functions that build sequences of values,`lazy-select-xml-tree-values` and `lazy-select-fn`, were decomposed into smaller functions which made it must easier to explore at a REPL. I'm not sure if I like the name `lazy-select-fn` or `lazy-cursor-fetch` better; thoughts on docstrings are most welcome.

### Miscellanea
- `xml-header` seemed like a nice abstraction; if nothing else, it makes
  reading `write-xml` a little easier
- extra logging

[bug]: https://www.pivotaltracker.com/story/show/134808839
[c3p0]: http://www.mchange.com/projects/c3p0/#basic_pool_configuration
[cursors]: https://www.postgresql.org/docs/9.4/static/plpgsql-cursors.html
[jdbc]: https://github.com/clojure/java.jdbc/blob/master/CHANGES.md


katrinamariehh:
🇨🇿

src/vip/data_processor/db/translations/util.clj
```diff
@@ -162,3 +163,21 @@
             (postgres/bulk-import postgres/xml-tree-values rows)
             (assoc :ltree-index (idx-fn)))
         ctx))))
+
+(defn transformer-with-conn
+  "Generates a transformation function that operates on a single
+  database connection? If `row-fn` is the sort of thing that requires a
```
tank157: Does it generate a transformation function that operates on a single db connection? Cause the docstring seems uncertain the way it's punctuated.

src/vip/data_processor/db/postgres.clj
```diff
@@ -358,77 +368,87 @@
...
-  (fn [& args]
-    (let [cursor-name (str (gensym "lazy_select_cursor"))
-          done? (atom false)
-          close-attempts (atom 0)
-          query (apply query-fn args)]
-      (korma/exec-raw
-       [(str "DECLARE " cursor-name " NO SCROLL CURSOR "
-             "WITH HOLD FOR "
-             query)])
+(defn from-cursor
```
tank157: Maybe `fetch-from-cursor` would be a clearer name here?

src/vip/data_processor/db/translations/util.clj
sduckett: Thanks, this docstring needs to be fixed.

PR #260, opened 2017-03-31
updated function to use precinct split polling locations

katrinamariehh:
This is a PR for [this bug](https://www.pivotaltracker.com/story/show/142810681).  The AR feed included data on the `v3_0_precinct_split_polling_locations` table rather than the `v3_0_precinct_polling_locations` table (which is totally valid).  The function that generates the polling location address error report was only using `v3_0_precinct_split_polling_locations`, so no data was generated for this AR feed.  

We updated the `v3_dashboard.polling_location_readable_report` function to use left joins and include joins to `v3_0_precinct_split_polling_locations` and `v3_0_precinct_splits` and then use those tables to join to `v3_0_precincts` and `v3_0_localities` (so you'll see that we join to each those two tables twice in the function, but using different source tables in each case).  Making them all left joins is necessary to work regardless of whether `v3_0_precinct_polling_locations` or `v3_0_precinct_split_polling_locations` is populated.

### DIFF
--- a/resources/migrations/20170331-00-polling-locations-include-precincts-splits.up.sql
+++ b/resources/migrations/20170331-00-polling-locations-include-precincts-splits.up.sql
@@ -4,7 +4,7 @@ as $$
 begin
 return query
 select distinct on (v.identifier)
-       l1.name, p1.name, pl.address_location_name, pl.address_line1,
+       coalesce(l.name, l0.name), coalesce(p.name, p0.name), pl.address_location_name, pl.address_line1,
        pl.address_line2, pl.address_line3, pl.address_city,
        pl.address_state, pl.address_zip, v.identifier
     from results r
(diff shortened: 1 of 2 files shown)


resources/migrations/20170331-00-polling-locations-include-precincts-splits.up.sql
```diff
@@ -0,0 +1,40 @@
+create or replace function v3_dashboard.polling_location_readable_report(pid text)
+returns table(locality_name text, precinct_name text, address_location_name text, address_line1 text, address_line2 text, address_line3 text, address_city text, address_state text, address_zip text, polling_location_id bigint)
+as $$
+begin
+return query
+select distinct on (v.identifier)
+       l1.name, p1.name, pl.address_location_name, pl.address_line1,
```
tie-rack: Seems to me (without running it, so please correct me if I'm wrong), but you're returning `p1.name`, which may be populated, but it could be `p.name` instead. Do we want something like `COALESCE(p.name, p1.name)`?

PR #261, opened 2017-04-12
Load ltree XML in parallel

sduckett:
When we're loading XML for a 5.1 feed, we turn each tag into a map in `path-and-values`. Since each of these maps exists in isolation, we can insert chunks in parallel! For a large feed, this cuts the runtime of this step in the pipeline by about 1/3; our large XML files were taking 30-45 minutes, but this brings them down to a slightly more reasonable 10-15 minutes. The cost of doing inserts in parallel is CPU load as the connection pool jumps to 15 simultaneous connections.

Credit goes to @tank157 for this suggestion.

### DIFF
--- a/src/vip/data_processor/validation/xml.clj
+++ b/src/vip/data_processor/validation/xml.clj
@@ -212,25 +212,39 @@
                         (:content node))]
      (apply concat attribute-entries child-entries))))
 
+(defn insert
+  "A function to insert to `xml_tree_values`. `vals` should contain
+  ltree objects rather than strings."
+  [vals]
+  (korma/insert postgres/xml-tree-values
+    (korma/values vals)))
+
+(defn paths->ltree
+  "Convert a fixed set of paths in `m` into Postgres objects and add an
+  `id`. Essentialls a composition of `postgres/ltreeify` and
+  `postgres/prep-for-insertion`."
+  [id m]
+  (-> m
+      (update :path postgres/path->ltree)
+      (update :parent_with_id postgres/path->ltree)
+      (update :simple_path postgres/path->ltree)
+      (assoc :results_id id)))
+
 (defn load-xml-ltree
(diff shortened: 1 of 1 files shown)


katrinamariehh:
I think I understand what's happening with the changes to the `validation/xml.clj`--but what's happening with the other two files?

sduckett:
> I think I understand what's happening with the changes to the validation/xml.clj--but what's happening with the other two files?

The stuff in `dev-src` is about exploring how different styles compare. I'm happy to remove it if you don't think it's relevant.

katrinamariehh:
It's fine!  I just wanted to make sure I understood.

src/vip/data_processor/validation/xml.clj
```diff
@@ -212,25 +212,39 @@
...
-                              (map (fn [path-map]
-                                     (-> path-map
-                                         (update :path postgres/path->ltree)
-                                         (update :parent_with_id postgres/path->ltree)
-                                         (update :simple_path postgres/path->ltree)
-                                         (assoc :results_id import-id)))
-                                   chunk))))]
-        (korma/insert postgres/xml-tree-values
-          (korma/values pvs))))
+      (doall (pmap insert
```
tank157: `doall` will cause the entire seq to reside in memory, is that a safe thing to do here? Not sure what the `insert` is returning here, but something to think about.

sduckett: `korma/insert` returns the last inserted row, so probably not an issue, but that might be an extraneous `doall`.

tank157: I had to use something like it when I did `pmap`, but I think maybe I used `dorun`

sduckett: It's subtle differences between the two, isn't it? I took a gander at memory use with Yourkit and noticed the old-gen space a little smaller, 100-200mb, with the total allocation a hair smaller.

After reading a bit more, `dorun` does seem more appropriate.

PR #262, opened 2017-04-12
new migrations to fix crosstab

katrinamariehh:
For [this card](https://www.pivotaltracker.com/n/projects/1106394/stories/143593837) a fix to the `v5_dashboard.locality_stats` function to deal with instances where data requested for the crosstab is unavailable.  Previously ,a missing  `Name` or `Type` would push the `id` out from the column where it's meant to be and we would not only lose that data on the locality table on the main feed page but then also lose the link to get more information about a specific locality.  This migration will properly align the data in the `locality` table on the main feed page so we don't lose that connection to the locality report.

### DIFF
--- a/resources/migrations/20170411-00-fix-locality-stats-crosstab.up.sql
+++ b/resources/migrations/20170411-00-fix-locality-stats-crosstab.up.sql
@@ -1,0 +1,206 @@
+drop function v5_dashboard.locality_stats (integer, text);
+
+create index xtv_paths on xml_tree_validations using gist (path);
+
+create function v5_dashboard.locality_stats(rid int, lid text)
+returns table(results_id int, id text, name text, type text, error_count int,
+              precinct_errors int, precinct_count int, precinct_completion int,
+              polling_location_errors int, polling_location_count int, polling_location_completion int,
+              street_segment_errors int, street_segment_count int, street_segment_completion int,
+              hours_open_errors int, hours_open_count int, hours_open_completion int,
+              election_administration_errors int, election_administration_count int, election_administration_completion int,
+              department_errors int, department_count int, department_completion int,
+              voter_service_errors int, voter_service_count int, voter_service_completion int)
+as $$
+  declare
+    locality record;
+    pbl ltree[];
+    precinct_polling_location_ids text[];
+    precinct_polling_locations ltree[];
+    locality_ea ltree[];
+    precincts text[];
+    street_segments ltree[];
+    precinct_errors int;
+    precinct_completion int;
+    street_segment_errors int;
+    street_segment_count int;
+    street_segment_completion int;
+    polling_location_errors int;
+    polling_location_count int;
+    polling_location_completion int;
+    hours_open_errors int;
+    hours_open_count int;
+    hours_open_completion int;
+    election_administration_errors int;
+    election_administration_count int;
+    election_administration_completion int;
+    department_count int;
+    error_count int;
+  begin
+    select ct.parent_with_id, ct.name, ct.type, ct.id
+    into locality
+    from crosstab('select xtv.parent_with_id, subpath(xtv.simple_path, -1) as element, xtv.value
+                   from xml_tree_values xtv
+                   where xtv.results_id = ''' || rid || '''
+                   and xtv.simple_path in (''VipObject.Locality.id'',
+                   ''VipObject.Locality.Name'',
+                   ''VipObject.Locality.Type'')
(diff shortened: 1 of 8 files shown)


resources/migrations/20170411-00-fix-locality-stats-crosstab.up.sql
```diff
@@ -0,0 +1,197 @@
...
+                   and xtv.simple_path in (''VipObject.Locality.id'',
+                   ''VipObject.Locality.Name'',
+                   ''VipObject.Locality.Type'')
+                   and xtv.path <@ (select subpath(xtv.parent_with_id, 0, 4)
+                                    from xml_tree_values xtv
+                                    where xtv.results_id = ''' || rid || '''
+                                      and xtv.simple_path = ''VipObject.Locality.id''
+                                      and xtv.value = ''' || locality_id || ''')
+                                    order by parent_with_id, element',
+                    'select unnest(ARRAY[''Name'', ''Type'', ''id''])' )
```
sduckett: This is the interesting line.
From the [tablefunc docs](https://www.postgresql.org/docs/9.4/static/tablefunc.html): _"Produces a "pivot table" with the value columns specified by a second query"_

sduckett:
### tl;dr
In addition to adding a [category query to crosstab](https://github.com/votinginfoproject/data-processor/pull/262/files#diff-4d54ef9270ab8cf7a09c45f5d40c49e1R54), this PR addresses some significant performance issues. A feed that once took 17 hours now takes about four and a half hours (that's loosely 25% of the original).

### Collecting Elements by Locality

In order to provide error reports scoped to localities, we collect the ltree `path` of any element that is connected to a locality (street segments, polling locations, precincts, election administrations, and electoral districts). This is a pretty expensive query for ad-hoc execution by the dashboard (given our current architecture), so we do this as [part of our processing pipeline](https://github.com/votinginfoproject/data-processor/pull/262/files#diff-70c2f35d9d9aa400f640505e8b2bdf03R31). You'll notice that this step has moved earlier in the pipeline; this is to let us [calculate the locality stats](https://github.com/votinginfoproject/data-processor/blob/master/src/vip/data_processor/errors/process.clj#L32) that run after validations a little more efficiently (for now, just the street segment errors, but more could be optimized with the same weird trick.)

### Finding Street Segment Errors

When checking the queries that calculate stats, we found the street segment errors were by far the slowest.

Migrations containing SQL functions don't lend themselves readily to diffs, so I'll put the relevant chunks side by side here. Originally, our query looked something like this:

```sql
select array_agg(subpath(xtv.parent_with_id, 0, 4))
into street_segments
from xml_tree_values xtv
where xtv.results_id = rid
    and xtv.simple_path = 'VipObject.StreetSegment.PrecinctId'
    and xtv.value in (select unnest(precincts));

select count(subpath(errors.path, 0, 4))
into street_segment_errors
from xml_tree_validations errors
where errors.results_id = rid
  and errors.path <@ street_segments
  and errors.path ~ '*.StreetSegment.*';
```

One thing to note is the order of the `where` clause at the end: first `results_id` is checked, then paths compared for relationship, and finally we check that it's a street segment error. The order in which predicates appear affects queries, so if you compare explain plans and actual runtime, you would notice that by moving `and errors.path ~ '*.StreetSegment.*'` to be the second predicate, runtime for a single locality would be cut in half without any indexes being altered.

Secondly, the path we're checking (`'*.StreetSegment.*'`) is a bit generous. We know from our schema where street segments live, so we can write a more specific path `'VipObject.0.StreetSegment.*'`. 

And finally, since we already know which street segment paths belong to any locality, we can rewrite this in terms of a value we've already calculated. You might wonder, _Don't we already have this in a variable named `street_segments`?_ Yes, we do, but that's a variable that is typed `ltree[]`, or an array of `ltree` paths. It turns out that checking whether one path is an ancestor of another takes about half as much time if the  descendant path is coming from an indexed table rather than an array. Neat, I guess.

```sql
with paths as (
  select unnest (paths) as path
  from v5_dashboard.paths_by_locality pbl
  where pbl.results_id = rid
      and pbl.locality_id = lid)
select count (v.path)
into street_segment_errors
from paths p, xml_tree_validations v
where v.results_id = rid
    and v.path ~ 'VipObject.0.StreetSegment.*'
    and v.path <@ p.path;
```

With these tweaks, I managed to turn a twenty minute query into a forty second query given the largest (by precinct count) locality of an Illinois feed. 

The biggest takeaway of this exercise is that generously matching paths are much slower when using the `ltree` extension; specify as much as you can for the best performance.

PR #263, opened 2017-04-24
Source & Election are required elements; verify their presence

sduckett:
Franklin noticed that we're [not validating the presence](https://www.pivotaltracker.com/story/show/144223187) of `Source` or `Election` elements, which should probably be fatal errors (my guess, since not knowing which election, on which date, a set of polling locations is for seems tricky). The 5.1 spec says [there must be one and only one](https://github.com/votinginfoproject/data-processor/blob/master/resources/specs/vip_spec_v5.1.2.xsd#L631), so that seems to fit with this being a fatal error.

This PR updates the validations to check for their presence. A feed missing both of these elements would have these two new errors shown on the dashboard like this:

screen shot 2017-04-24 at 4 10 44 pm

### DIFF
--- a/src/vip/data_processor/validation/v5/election.clj
+++ b/src/vip/data_processor/validation/v5/election.clj
@@ -17,10 +17,14 @@
         election-count (-> result
                            first
                            :election_count)]
-    (if (> election-count 1)
-      (errors/add-errors ctx :fatal :election "VipObject.0.Election" :count
-                         :more-than-one)
-      ctx)))
+    (cond
+      (zero? election-count) (errors/add-errors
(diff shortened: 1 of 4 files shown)


katrinamariehh:
So just to confirm--previously we were only checking if there was more than one source or election, now we make sure there are not zero sources or elections as well as checking that there is not more than one of either of them.

sduckett:
> So just to confirm--previously we were only checking if there was more than one source or election, now we make sure there are not zero sources or elections as well as checking that there is not more than one of either of them.

That's correct; there are two error cases to consider here.

katrinamariehh:
Cool!  LGTM

PR #264, opened 2017-05-05
Replace any return char with \n in 5.1 text blocks

tank157:
Fixes the formfeed issue identified in [this bug](https://www.pivotaltracker.com/story/show/144406851). Verified that before this fix the provided CSV feed fails to run in `xmllint` with an invalid char. After, the feed still fails to be valid according to `xmllint`, but for issues of structure (missing `<Name>` elements for one).

I applied the `clean-text` function to any 5.1 data-spec fields that seemed like they could be larger blocks of text where newlines are likely to appear.


katrinamariehh:
👍🏻

PR #265, opened 2017-05-12
Remove xml tree values refs

katrinamariehh:
# The what*
The pull request executes the results of the research outlined [here](https://www.pivotaltracker.com/story/show/139318429).  It is intimately connected to [this Metis PR](https://github.com/votinginfoproject/Metis/pull/362).

# The why*
Because the `xml_tree_values` table holds mostly raw data and is the largest table in our database, we decided to figure out how to make all of our necessary summary data accessible after processing is complete without touching `xml_tree_values`.  This would let us delete all of a feed's rows in `xml_tree_values` at the end of each processing run instead of retaining raw data when it is no longer useful.  A few weeks ago the production database ran of space, so this has become more of a priority as more states are submitting 5.1 feeds which necessarily take up more database space.

# The how*
## Add `parent_element_id` to `xml_tree_validations`

Retrieving the ID of the parent element is the primary reason that queries or functions join to `xml_tree_values` after processing.  To remove that dependency, I added `parent_element_id` as a column on `xml_tree_validations`. [(00)](https://github.com/votinginfoproject/data-processor/compare/remove-xml-tree-values-refs#diff-dc5d513e8c2ad15c29bf8d41be226f6d) This also required a migration to populate that field for `xml_tree_validations` values currently in the database.  [(01)](https://github.com/votinginfoproject/data-processor/compare/remove-xml-tree-values-refs#diff-38ceee07bb047dfe1cc33262a8d4d8e8)  

The only function on the dashboard that runs after processing and joins to `xml_tree_values` is the `v5_dashboard.locality_error_report`, so it would also need to be rewritten to access `parent_element_id` on `xml_tree_validations` in place of the join.  [(02)](https://github.com/votinginfoproject/data-processor/compare/remove-xml-tree-values-refs#diff-c61d2973e8912b8b6991de8091c73357) While confirming that that the changes were working as expected it became clear that for very large feeds this function was taking upwards of 60 seconds to return results.  Rewriting a small part of that function improved it's performance for large feeds significantly (the Louisiana feed which could take up to 90 seconds to run this function for any given locality can now run in less than a second).

## Rewrite pipeline functions to include `parent-element-id` when adding errors

The second step was to rewrite functions in the processing pipeline to ensure that when errors are written out to `xml_tree_validations` that they now include `parent_element_id`.  This required locating where in the v5 validations process the `add-errors` function is called and editing the call to include this new value.  This was narrowed down to only the `add-errors` calls in the `/validations/v5` directory—all other `add-errors` calls are either relevant only to v3 feeds which we are unconcerned with or are validations for non-xml features which necessarily do not have `parent_element_id`s (and therefore default to `null` on `xml_tree_validations` as they would have in the previous schema).  Initially I had hoped to have two versions of `add-errors` with different function signatures but the same name, but the use of rest args means the functions must have two different names, thus `v5-add-errors` and a new record `V5ValidationError`.

## Delete values from `xml_tree_values` at the end of a processing run

The terminology used originally was that we would truncate `xml_tree_values` at the end of processing to drop values for a feed and decrease the database size, but because we have three workers doing processing potentially at the same time we instead will specifically [delete the values for a given feed from `xml_tree_values`](https://github.com/votinginfoproject/data-processor/compare/remove-xml-tree-values-refs#diff-1912fbc92d6ff1c6c1aa3801737b200e) at [the end](https://github.com/votinginfoproject/data-processor/compare/remove-xml-tree-values-refs#diff-5184e1c12d8253c58d9e73115351cc0d) of the processing pipeline.

# Final thoughts
We did initially explore the possibility of interrupting at the time of writing to the database to add this new value (i.e. when the rows are mapped in `/db/postgres.clj`) or adding a postgres function at the end of the run that would populate the column (such as the one included in the migration to populate the column for existing data), but in the end we decided that changing `add-errors` was less side-effect-y and would be a clearer implementation in the long run.  It should be noted, though, that in the end @sduckett and I both think that this implementation doesn't quite *feel* right to us, though neither of us can say quite what the problem is.

I did *not* include a migration to drop values from `xml_tree_values` for data currently in the database, but that may need to be a part of this.  The deployment of these changes and the [Metis PR](https://github.com/votinginfoproject/Metis/pull/362) will need to be considered together to make sure all the data that needs to be accessed is available or otherwise some down time will need to be scheduled.

*why yes I _have_ been watching a lot of dumb detective shows recently, why ever would you ask?


resources/migrations/20170428-01-add-parent-element-id-to-xml-tree-validations.up.sql
```diff
@@ -0,0 +1,12 @@
...
+        v.error_type, v.error_data
+ FROM xml_tree_validations v
+ LEFT JOIN xml_tree_values x
+        ON x.path = subpath(v.path,0,4) || 'id' AND x.results_id = v.results_id
+ INNER JOIN results r ON r.id = v.results_id)
+
+update xml_tree_validations v
+set parent_element_id = values.identifier
+from values
+where v.path = values.path;
```
sduckett: I'm curious why there are two migrations here: one to add a column and one to populate the column (I would have probably combined them to begin with). Is there a case where we might need to have the column available but without any data, e.g., rolling back only the value portion?

Also, I can imagine this update taking a while to apply, which might be interesting around deployment time.

src/vip/data_processor/db/postgres.clj
```diff
@@ -301,6 +301,14 @@
    ["select v5_dashboard.populate_elections_table(?)" [import-id]])
   ctx)
 
+(defn delete-from-xml-tree-values
+  [{:keys [import-id] :as ctx}]
+  (log/info "Dropping from xml_tree_values")
+  (korma/exec-raw
```
sduckett: If you enjoy the DSL nature of Korma, you might consider [`korma.core/delete`](http://korma.github.io/Korma/korma.core.html#var-delete) instead of using `korma.core/exec-raw`.

src/vip/data_processor/errors.clj
```diff
@@ -14,6 +14,18 @@
 
   ctx)
 
+(defrecord V5ValidationError [ctx severity scope
+                                identifier error-type error-value parent-element-id])
+
+(defn v5-add-errors [{:keys [errors-chan] :as ctx}
```
sduckett: Not sure it makes any difference, but maybe `add-v5-errors`, with the verb first? (_adjective-verb-noun_ feels weird to me here.)

src/vip/data_processor/validation/v5/candidate.clj
```diff
@@ -31,9 +40,18 @@
                                  :simple_path (postgres/path->ltree "VipObject.Candidate.PostElectionStatus")}))
         invalid-statuses (remove valid-post-election-status? statuses)]
     (reduce (fn [ctx row]
-              (errors/add-errors ctx
-                                 :errors :candidates (-> row :path .getValue) :format
-                                 (:value row)))
+              (let [parent-element-id (->(korma/exec-raw
```
sduckett: It looks like each of the validations binds `parent_element_id` to a function of `row` and `import-id` like this:

```clojure
(let [parent-elemnt-id (-> (korma/exec-raw
                             (:conn postgres/xml-tree-values)
                             ["SELECT value FROM xml_tree_values
                               WHERE path = subpath(text2ltree(?),0,4) || 'id'
                                 and results_id = ?" [(-> row :path .getValue) import-id]]
                             :results)
                           first
                           :value)])
```
Maybe this would be worth extracting to a utility function? Also, does the order of the `where` clause affect query plans? Most of our indexes have `results_id` as their first value; I remember finding a few cases where order was pretty important (and others where it was totally irrelevant).

src/vip/data_processor/errors.clj
tank157: Yeah, I'd agree that `verb-adjective-noun` seems better order to me.

src/vip/data_processor/validation/v5/candidate.clj
tank157: A utility function for this would be nice, as I'll have to rework some of the Start/End House Number validations to use the `v5` error structure, and thus need to get the parent id too.

katrinamariehh: I did think about making it a utility function, but the way we get the `parent-element-id` varies depending on what else we have access to in the function i.e. sometimes I have an ltree `path`, sometimes it's a string that has to be cast to an ltree `path` so the utility function would have to be two functions depending on that or it would have to determine which one it has within the function.  What sounds best here?

resources/migrations/20170428-01-add-parent-element-id-to-xml-tree-validations.up.sql
katrinamariehh: I think it just seemed to make sense to me at the time?  I'm not sure that there's an explicit reason, I'm happy to change it.

src/vip/data_processor/validation/v5/util.clj
```diff
@@ -193,9 +227,18 @@
             elements (elements-at-simple-path import-id simple-path)
             invalid-elements (remove (comp valid? :value) elements)]
         (reduce (fn [ctx row]
-                  (errors/add-errors ctx error-severity error-root-element
-                                     (-> row :path .getValue) error-type
-                                     (:value row)))
+                  (let [parent-element-id (->(korma/exec-raw
```
tie-rack: Can `parent-element-id` here and in places above use the `get-parent-element-id` function?

katrinamariehh: Yep!  Except for `build-xml-tree-value-query-validator` which has an ltree path instead of something that needs `text2ltree` I can use that function again (the function was built to expect a string and to have to use `text2ltree` because in every case except for this function that is what we have).  Thanks!

PR #266, opened 2017-05-16
validate start and end house numbers

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/144313917)

Implements the backend portion of the start/end house number validation.


src/vip/data_processor/validation/v5/street_segment.clj
```diff
@@ -3,6 +3,7 @@
             [vip.data-processor.db.postgres :as postgres]
             [vip.data-processor.validation.v5.util :as util]
             [clojure.string :as str]
+            [clojure.edn :as edn]
```
sduckett: Looks like this is an extraneous require?

tank157: Good catch, it was left over from a previous implementation where I was using the edn reader to parse the value.

sduckett:
👍

PR #267, opened 2017-06-12
added function to get early vote site address error report

katrinamariehh:
For [this card](https://www.pivotaltracker.com/story/show/145550577) and related to [this PR](https://github.com/votinginfoproject/Metis/pull/370)--a database function to generate an address-specific error report for Early Vote Sites.  It grabs all early vote sites with an address-related error and groups them together i.e. if your Early Vote Site is missing both a zip and a state, that site will have it's own line with those two pieces missing.

### DIFF
--- a/resources/migrations/20170609-00-early_vote_address_report.down.sql
+++ b/resources/migrations/20170609-00-early_vote_address_report.down.sql
@@ -1,0 +1,1 @@
+drop function v3_dashboard.early_vote_location_address_report(text);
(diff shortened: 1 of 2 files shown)


katrinamariehh:
Maybe hold off for a bit--I did not actually check the example report on the card to make sure it's doing what Maria expected.

katrinamariehh:
This is also good to go--added `locality_name` back to the report here as well

katrinamariehh:
I'm hesitant to add in an index if it's not obviously necessary--as it is the things I've been testing run in 2.2 ms and I don't know that we have any examples of 3.0 feeds that we could even test against in a concern for speed.

tank157:
It might be ok to leave the index off, but indexes are usually applied so that things run as well in production where there's a lot of data as they do in dev. That said, it might be that this particular part of the data set isn't ever going to be that big, since there may only be dozens of early vote sites compared to millions of street segments. And there aren't likely to be that many elections with early vote sites as they aren't that common yet.

katrinamariehh:
There are only 400 early vote sites in the production database, I think we're fine to leave them off.

PR #271, opened 2017-07-17
added performance improvements to feed_localities and locality_stats functions

katrinamariehh:
To improve the performance of the `feed_localities` and `locality_stats` functions I added code to the `feed_localities` function to generate a temporary table of the street segment validations for the feed it is running with and then altered the `locality_stats` function to get its `street_segment_errors` from the temporary table instead of from `xml_tree_validations`.  I also added an index and evaluated the table (so it would pick up the index, see [this stack overflow post](https://dba.stackexchange.com/a/153439)).  This step for a Louisiana feed was previously taking 4.5 hours and is now down to 6 minutes.  Because the changes are in migration files it is difficult to see the differences, but comparing the `up` and `down` files should be sufficient.  I also haven't yet but should deploy it to staging and test it out there as well.


resources/migrations/20170707-00-improve-locality-stats-function.up.sql
```diff
@@ -0,0 +1,205 @@
...
+    polling_location_completion := v5_dashboard.completion_rate(polling_location_count, polling_location_errors);
+
+    with paths as (
+      select unnest (paths) as path
+      from v5_dashboard.paths_by_locality pbl
+      where pbl.results_id = rid
+        and pbl.locality_id = lid)
+    select count (v.path)
+    into street_segment_errors
+    from paths p, v5_street_segment_validations v
```
katrinamariehh: This is the change where it requests from the temp table instead of from xml_tree_validations; there's a 2-line difference in the file because it changes the source table in this instance and it no longer needs the where clause to narrow down the path being a street segment path (both of those things are taken care of in the temp table creation process).

resources/migrations/20170707-00-improve-locality-stats-function.down.sql
```diff
@@ -0,0 +1,206 @@
...
+    polling_location_completion := v5_dashboard.completion_rate(polling_location_count, polling_location_errors);
+
+    with paths as (
+      select unnest (paths) as path
+      from v5_dashboard.paths_by_locality pbl
+      where pbl.results_id = rid
+        and pbl.locality_id = lid)
+    select count (v.path)
+    into street_segment_errors
+    from paths p, xml_tree_validations v
```
katrinamariehh: This is the portion getting replaced in the new function.

PR #272, opened 2017-07-25
Stop indexing values that may be huge

tie-rack:
Removes `value` from `xtv_i18n_text_idx` and specifies paths in the index made by `v5_dashboard.populate_locality_table` to not include many paths that should not be indexed.

### DIFF
--- a/resources/migrations/20170725-00-remove-value-from-xtv_i18n_text_idx.down.sql
+++ b/resources/migrations/20170725-00-remove-value-from-xtv_i18n_text_idx.down.sql
@@ -1,0 +1,5 @@
+DROP INDEX IF EXISTS xtv_i18n_text_idx;
+
+CREATE INDEX xtv_i18n_text_idx
+ON xml_tree_values (results_id, path, value)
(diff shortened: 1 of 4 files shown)


katrinamariehh:
LGTM--can you take a look @tank157 and see what you think?  It looks like the main differences are [here](https://github.com/votinginfoproject/data-processor/pull/272/files#diff-dde74f37525af50190d43cf3dd49f82eR32) to [here](https://github.com/votinginfoproject/data-processor/pull/272/files#diff-35b612aa3cab8f641bcbd04e7d024302R32) where the `simple_path`s for a more specific set of `value`s is defined for creating the index.  I ran it locally and didn't see the index-related error showing up anymore.

tank157:
Ah, ok, it took a while to tease out, it looks like the major change is teasing out [this line](https://github.com/votinginfoproject/data-processor/pull/272/files#diff-dde74f37525af50190d43cf3dd49f82eR45). Instead of matching absolutely everything under, say, BallotMeasureContest, you're only matching on certain values that are known to be small.

PR #273, opened 2017-08-28
Upgrade squishy

tank157:
Reviving this older branch so we can take advantage of the visibility-timeout feature to prevent any super long running feeds from effectively DOSing the data-processor nodes.

Setting the timeout to 12 hours initially to match the current setting on the queues.

### DIFF
--- a/resources/config.edn
+++ b/resources/config.edn
@@ -4,7 +4,8 @@
             :processed-bucket #resource-config/env "VIP_DP_S3_PROCESSED_BUCKET"}
        :sqs {:region #aws/region #resource-config/env "VIP_DP_SQS_REGION"
              :queue #resource-config/env "VIP_DP_SQS_QUEUE"
-             :fail-queue #resource-config/env "VIP_DP_SQS_FAIL_QUEUE"}}
+             :fail-queue #resource-config/env "VIP_DP_SQS_FAIL_QUEUE"
+             :visibility-timeout 43200}}
  :postgres {:host #resource-config/env "DB_PORT_5432_TCP_ADDR"
             :port #resource-config/env "DB_PORT_5432_TCP_PORT"
             :user #resource-config/env "DB_ENV_POSTGRES_USER"
--- a/project.clj
+++ b/project.clj
@@ -10,9 +10,10 @@
                  [clj-time "0.12.2"]
                  [clj-aws-s3 "0.3.10" :exclusions [joda-time]]
                  [com.climate/clj-newrelic "0.2.1"]
-                 [democracyworks/squishy "2.0.0" :exclusions [joda-time
-                                                              org.slf4j/slf4j-simple
-                                                              org.slf4j/slf4j-api]]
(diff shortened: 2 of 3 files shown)


resources/config.edn
```diff
@@ -4,7 +4,8 @@
             :processed-bucket #resource-config/env "VIP_DP_S3_PROCESSED_BUCKET"}
        :sqs {:region #aws/region #resource-config/env "VIP_DP_SQS_REGION"
              :queue #resource-config/env "VIP_DP_SQS_QUEUE"
-             :fail-queue #resource-config/env "VIP_DP_SQS_FAIL_QUEUE"}}
+             :fail-queue #resource-config/env "VIP_DP_SQS_FAIL_QUEUE"
+             :visibility-timeout 43200}}
```
cap10morgan: You don't want to match the queue setting. You want to be well under it. Otherwise it's kind of a race condition to who gets there first. If this wins, the visibility timeout will get reset and all will be well. If SQS wins the message will timeout and get re-queued.

In general a very short timeout is what you want here. One measured in tens-of-seconds or minutes.

tank157: In that case the documentation in the README for squishy is misleading. The example it gives says there's a 300 second timeout on the SQS queue and then proceeds to use a value of `{:visibility-timeout 300}` as the options. It wondered about that, and looked at the code. What it seemed to be doing was picking a time half-way between the visibility timeout and then resetting it then. Is that not the case?

cap10morgan: Ah, OK. It's been awhile. I was remembering the algorithm and not the code I guess. Yeah, so I guess this is intended to match the queue's setting after all. But more broadly this feature is intended to obviate the need to customize the visibility timeout at all. That way if you ever point this at a new SQS queue and don't change the visibility timeout from the default, things will just work. So you may want to consider going back to the default visibility timeout everywhere.

tank157: We can consider it. Or perhaps setting it at a more reasonable setting for the majority of feeds. There is a potential cost for every SQS action (every API call counts as an action), though it seems they are free until we've made 1 million of them. If we had feeds processing processing near constantly on all three nodes it still wouldn't hit 1 million requests per month at a default of 30 seconds but would get to like 25% of that (not to mention the initial SQS action to post the message). So maybe some compromise, like 10 minutes would be good.

katrinamariehh:
So will this actually prevent the resubmission of long-running feeds that we've been seeing or is this the first step to get that?  Everything looks fine to me, I just want to make sure I understand what we're going for.

tank157:
This in theory will prevent them from re-submitting, yes. It'll be hard to demo/show in practice unless we want to do a couple of things, like reduce the visibility timeout to some smaller value, change the config value in `config.edn` here to match, and then kick off a feed that will take longer than that.

katrinamariehh:
I think it's probably fine to go ahead and merge--it's a relatively rare occurrence, so we'll just want to keep an eye on things, I think.

PR #275, opened 2017-09-12
Polling locations by type

katrinamariehh:
For [this card](https://www.pivotaltracker.com/story/show/150649272) and related to [this](https://github.com/votinginfoproject/Metis/pull/390) Metis PR, I edited the function that generates the data for the `v5_statistics` table to ignore polling location data which instead is populated by a new function `v5_dashboard. polling_locations_by_type` so that we can break out the data by polling location type (i.e. regular, drop box, or early voting).

The main functionality is in the migration that creates the function; there is some repetitiveness at the end when the completion is generated for each type.  If you have any suggestions about fixing that I would love to hear it!


src/vip/data_processor/db/tree_statistics.clj
```diff
@@ -10,8 +10,10 @@
   (->> postgres/v5-statistics
        postgres/column-names
        (filter #(str/ends-with? % "_count"))
+       (remove #(str/starts-with? % "polling_location"))
        (map #(str/replace % #"_count" ""))
-       (map t-util/column->xml-elment)))
+       (map t-util/column->xml-elment)
+       (remove #(str/starts-with? % "PollingLocation"))))
```
tank157: This function is somewhat opaque. Not that it wasn't already before, but it seems confusing to have to remove Polling Locations twice. I wonder if there's some way we can make this clearer? Tests could help, but they'd have to be the integration style tests that rely on a db. Maybe just a doc string?

katrinamariehh: I'm not sure if they both need to be there, let me check on this and see.

PR #276, opened 2017-09-19
Duplicate polling place warnings

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/150649639)

Runs a check for polling location ids in both localities and precincts, and adds a warning with all of the ids if any are found.

### DIFF
--- a/src/vip/data_processor/errors.clj
+++ b/src/vip/data_processor/errors.clj
@@ -5,9 +5,16 @@
 (defrecord ValidationError [ctx severity scope
                             identifier error-type error-value])
 
-(defn add-errors [{:keys [errors-chan] :as ctx}
-                  severity scope identifier error-type
-                  & error-data]
+(defn add-errors
+  "Adds an error to the validations.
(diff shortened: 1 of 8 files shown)


src/vip/data_processor/errors.clj
```diff
@@ -17,9 +24,13 @@
 (defrecord V5ValidationError [ctx severity scope
                                 identifier error-type error-value parent-element-id])
 
-(defn add-v5-errors [{:keys [errors-chan] :as ctx}
-                     severity scope identifier error-type parent-element-id
-                     & error-data]
+(defn add-v5-errors
+  "Same error options as add-errors, also with a parent-element-id which
```
katrinamariehh: I think this could probably be explained better--the reason why it exists separately from `add-errors` is to specifically associate the `parent-element-id` which is required for all of the error reports; adding it here allows us to generate error reports without accessing `xml_tree_values` which gets deleted at the end of each feed processing run.

tank157:
@katrinamariehh Can you check the change I made and let me know if this needs more to be accepted?

PR #278, opened 2017-11-28
Add validation check for invalid xml characters

tank157:
After several dead-ends and wild goose chases, I finally tracked down a sensible, workable way to do this. While the [Pivotal Card](https://www.pivotaltracker.com/story/show/152411377) mentions non UTF-8 characters being the problem, it's actually valid UTF-8 characters that are invalid in XML. So this check will look at all the string values and check them for any invalid XML characters. For the test I added, I used the actual bad file from the card, stripped it down and removed identifying information, and the test passes in that it finds the expected error because of the bad character. It should help someone find the problem, but in the case of control characters like this, it may still be difficult as many/most editors will hide it. Still, progress!


katrinamariehh:
💯

PR #280, opened 2018-03-15
Candidate address validation

ddellacosta:
This adds a check to confirm we have all the fields required when we have a [`filed_mailing_address`](https://github.com/votinginfoproject/data-processor/blob/bce02bcd0e8be67c6211d0edcf3f2d92a3d8b26d/resources/specs/vip_spec_v3.0.xsd#L206) for a candidate, which is an instance of a [`simpleAddressType`](https://github.com/votinginfoproject/data-processor/blob/bce02bcd0e8be67c6211d0edcf3f2d92a3d8b26d/resources/specs/vip_spec_v3.0.xsd#L336). This specifies that we should have, at the least, `line1`, `city`, `state`, and `zip`. While this does not mean we are now confirming the presence of these four fields for _any_ `simpleAddressType` in a feed, we are for a `candidate`, at least.

Correspondingly, while the card (https://www.pivotaltracker.com/story/show/151256241) only highlighted the bug where the ingestion breaks because of a missing `city` field for a candidate's address, after checking with Franklin he agreed that this we should be checking for all four of these fields.

Finally, I've set the severity to `:critical` after discussion with Franklin as well. I've also confirmed that the validation errors I've added here are showing up without any further work in the dashboard, both in the validation errors export CSV as well as in the appropriate places on the dashboard UI itself.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,7 @@ processor:
 rabbitmq:
   image: rabbitmq:3.6.5-management
   ports:
+    - 45672:5672
     - 55672:15672
   hostname: qa-engine-rabbit
 db:
--- a/src/vip/data_processor/validation/db/v3_0.clj
+++ b/src/vip/data_processor/validation/db/v3_0.clj
@@ -1,12 +1,15 @@
 (ns vip.data-processor.validation.db.v3-0
-  (:require [vip.data-processor.validation.db.v3-0.admin-addresses :as admin-addresses]
-            [vip.data-processor.validation.db.v3-0.precinct :as precinct]
-            [vip.data-processor.validation.db.v3-0.street-segment :as street-segment]
-            [vip.data-processor.validation.db.v3-0.jurisdiction-references :as jurisdiction-references]
-            [vip.data-processor.validation.db.v3-0.fips :as fips]))
+  (:require
+   [vip.data-processor.validation.db.v3-0.admin-addresses :as admin-addresses]
+   [vip.data-processor.validation.db.v3-0.candidate-addresses :as candidate-addresses]
+   [vip.data-processor.validation.db.v3-0.fips :as fips]
+   [vip.data-processor.validation.db.v3-0.jurisdiction-references :as jurisdiction-references]
+   [vip.data-processor.validation.db.v3-0.precinct :as precinct]
+   [vip.data-processor.validation.db.v3-0.street-segment :as street-segment]))
 
 (def validations
   [admin-addresses/validate-addresses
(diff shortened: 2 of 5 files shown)


ddellacosta:
@tank157 Good call, will add.

sduckett:
> While this does not mean we are now confirming the presence of these four fields for any simpleAddressType in a feed, we are for a candidate, at least.

Is this a validation that should be added for the other elements that use `simpleAddressType`?

tank157:
@sduckett It does look like the XSD puts the restrictions for a simpleAddress on the type itself, so there could be a case for that. The question would be if we implemented similar validations in the other places it gets used: `polling_location`, `early_vote_site`, `polling_locationType` (what is that?), and `election_administration`.

ddellacosta:
@sduckett @tank157 

I think the problem is that we don't have anywhere that we have isolated `simpleAddressType` as a value itself where we could do validation easily. As Troy points out it seems like we'd have to either implement that for all the other types that use `simpleAddressType` (and I think we already have for `election_administration`, in fact I based this PR off of that somewhat), or find a way to do a more sophisticated validation at the `data-spec` level. Right now that code only seems to handle checking whether individual values are present or not, so that seems like a significant overall of that code to get this working in a less verbose way.

As it stands, I can mention it in the card (that we aren't necessarily validation `simpleAddressType`s everywhere we could), but I'm reticent to do much more work here as it's not obvious if we care about it or not at this point, on the level of what comes out of the feed and ingestion by Google.

ddellacosta:
@sduckett Let me also add, if I'm wrong in my analysis of the code let me know--maybe there _is_ somewhere that would let us evaluate `simpleAddressType` addresses as `simpleAddressType` addresses?

sduckett:
> I'm reticent to do much more work here as it's not obvious if we care about it or not at this point...
@ddellacosta

ddellacosta:
@sduckett Sorry I'm not following--what are you suggesting by quoting me there?

sduckett:
@ddellacosta I'm saying you don't need to do anything; I was just thinking out loud that we might come back to this idea later on when we learn something new about how Google does things.

ddellacosta:
@sduckett Gotcha! Yeah, that had occurred to me too--I suspect that we will, in fact, have to revisit this later precisely because of stuff dying at the feed ingestion stage.

In any case, thanks for your comment, it was definitely a good question and worth thinking through. My fantasy is to some day automate at least some aspects of validation based on simply processing the XSD, but preliminary experiments have shown me that it won't be trivial (guessing I was not the first to attempt this either...).

PR #281, opened 2018-03-23
refactor bulk-import-and-validate-csv into a number of smaller functions

ddellacosta:
...and add some docstrings to better document the intermediate steps in importing a CSV.

As I've been going through `data-processor` I've found many areas that are hard to understand, but as opposed to other parts of the code, this function was a place where I thought I could give us a quick refactoring win. This refactoring should not change the logic of the process, the goal here was simply to take an 80-line, hard to understand function and refactor it into a set of smaller, easy-to-understand functions, each with its own obvious purpose. This should also make the flow of data through the CSV import process more clear so that augmenting this or refactoring it more substantially at a later date becomes simpler.

Any further refactoring comments are very welcome--I really hope you let me know where things are still unclear. Ideally, by the time I'm ready to merge this PR in, I hope to have replaced the original function with some code that anyone on our team can read through and get the gist of pretty quickly.

### DIFF
--- a/src/vip/data_processor/validation/csv.clj
+++ b/src/vip/data_processor/validation/csv.clj
@@ -52,7 +52,7 @@
   "Uses the supplied `format-rules` to process a given row. If there
   is a mismatch in the size of the columns in the row and the columns
   specified by the header, records an error. Increments the
-  line-number and returns the context ctx with the formatted row
+  line-number and returns the context `ctx` with the formatted row
   merged in."
   [{:keys [filename table] :as data-spec}
    {:keys [line-number headers headers-count format-rules] :as file-ctx}
@@ -82,11 +82,11 @@
 (defn process-row-chunks
   "Processes all the map rows in a chunk via `process-row` with the
   supplied format-rules (see `data-spec/create-format-rules` in
-  `do-bulk-import-and-validate-csv`) transforms (see
+  `do-bulk-import-and-validate-csv`) and transforms (see
   `data-spec/translation-fns` in
   `do-bulk-import-and-validate-csv`). Inserts the row into the
   database, and records any UNIQUE constraint-failing SQLExceptions as
-  errors. Returns the context ctx. If the supplied chunk is empty,
+  errors. Returns the context `ctx`. If the supplied chunk is empty,
   simply returns the context as-is."
   [{:keys [columns filename] :as data-spec}
    {:keys [transforms column-names headers sql-table line-number] :as file-ctx}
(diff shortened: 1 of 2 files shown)


src/vip/data_processor/validation/csv.clj
```diff
@@ -41,100 +42,183 @@
...
+       (map str/lower-case)))
+
+(defn required-header-names
+  "Given a collection of column maps, returns cleaned (of non word
+  characters), lower-cased string names (:name) of all the columns
+  which are required (:required)."
+  [columns]
+  (-> (comp (filter :required)
+            (map :name)
+            (map clojure.string/lower-case))
```
tank157: `clojure.string` is aliased to `str` already, can we change this to `str/lower-case`?

src/vip/data_processor/validation/csv.clj
```diff
@@ -41,100 +42,183 @@
...
-       first))
+  (->> reader .readLine csv/read-csv first))
+
+(defn warn-if-extraneous-headers
+  "Records an error if headers are present in the file that do not
+  correspond to any in the data-spec (`column-names`)."
+  [ctx {:keys [table] :as data-spec} headers column-names]
+  (let [extraneous-headers (seq (set/difference (set headers) (set column-names)))]
+    (if extraneous-headers
+      (apply errors/add-errors
```
tank157: Not a problem, just curious of the choice of `apply` here instead of directly calling the `errors/add-errors` function?

src/vip/data_processor/validation/csv.clj
```diff
@@ -41,100 +42,183 @@
...
+  specified by the header, records an error. Increments the
+  line-number and returns the context `ctx` with the formatted row
+  merged in."
+  [{:keys [filename table] :as data-spec}
+   {:keys [line-number headers headers-count format-rules] :as file-ctx}
+   ctx row]
+  (swap! line-number inc)
+  (let [row-map (zipmap headers row)
+        row-count (count row)
+        ctx (data-spec/apply-format-rules format-rules ctx row-map @line-number)]
```
tank157: I don't think/know that we'd ever parallelize this, but wonder if it would be wise to add a `current-line @line-number` to the let form since it's dereferenced in multiple spots, and then if we did parallelize it, we'd be certain of a consistent line-number value for the entire execution of the function.

src/vip/data_processor/validation/csv.clj
ddellacosta: No idea why I did that!! Will fix, absolutely.

src/vip/data_processor/validation/csv.clj
ddellacosta: Oh, just because `extraneous-headers` is a collection I guess...it was like that before so I didn't bother tweaking it.

src/vip/data_processor/validation/csv.clj
ddellacosta: Honestly I'd rather factor out the atom entirely if possible...seems like something we could introduce as a local variable, I just didn't bother trying to tackle that at this point.

PR #282, opened 2018-03-23
fix too-strict candidate address validation

ddellacosta:
The candidate validation I added for https://github.com/votinginfoproject/data-processor/pull/280 was too strict, as it wouldn't allow candidates with _no_ addresses through, even though the [`filed_mailing_address` for a candidate is optional](https://github.com/votinginfoproject/data-processor/blob/b22a60dd0ef29a5c857528dda918e1b14c2686a6/resources/specs/vip_spec_v3.0.xsd#L206).

This fixes the over-zealousness of the validation, and adds a test to confirm that a candidate with no address is **not** marked as invalid, along with continuing to correctly mark those candidates with addresses but missing required fields as invalid (and those with all required fields are valid).

### DIFF
--- a/src/vip/data_processor/validation/db/v3_0/candidate_addresses.clj
+++ b/src/vip/data_processor/validation/db/v3_0/candidate_addresses.clj
@@ -10,22 +10,22 @@
     :filed_mailing_address_state
     :filed_mailing_address_zip})
 
-(defn invalid-address?
+(defn valid-address?
   [address]
   (let [address' (dissoc address :id)]
     (cond
       ;; if all address fields are empty then is valid
       (not (seq (remove str/blank? (vals address'))))
-      false
+      true
 
       ;; if there is an address with any required fields blank, then invalid
       (->> ((apply juxt required-address-fields) address')
            (some str/blank?))
-      true
+      false
 
       ;; Otherwise we have a valid address with all required fields
       :else
-      false)))
+      true)))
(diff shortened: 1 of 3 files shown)


src/vip/data_processor/validation/db/v3_0/candidate_addresses.clj
```diff
@@ -4,6 +4,29 @@
...
+  (let [address' (dissoc address :id)]
+    (cond
+      ;; if all address fields are empty then is valid
+      (not (seq (remove str/blank? (vals address'))))
+      true
+
+      ;; if there is an address with any required fields blank, then invalid
+      (->> ((apply juxt required-address-fields) address')
+           (some str/blank?))
+      false
```
gws: Should this guard be tested first? Otherwise, barring some execution model where you re-validate addresses, it seems that this will be run after the first guard testing for all-blank-fields, and an entirely blank address with required fields would pass the first test in the `cond`.

ddellacosta: Shit! I saw this too late. Let me review and confirm my logic here.

ddellacosta: Okay, I think we're good, but thanks for taking another look at this Gordon--never hurts to double-check the logic.

PR #285, opened 2018-04-19
replace cleanup function with one that works with just results

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/156922010)

Replaces the cleanup function with one that will catch 5.1 feeds. Once we started deleting from `xml_tree_values` when processing was over, this couldn't clean up 5.1 feeds anymore. I tested this locally, processed 7 copies of the same feed, and it figured out which 2 to delete.

### DIFF
--- a/resources/migrations/20180419-fix-cleanup.up.sql
+++ b/resources/migrations/20180419-fix-cleanup.up.sql
@@ -1,0 +1,22 @@
+create schema if not exists cleanup;
+
+create or replace view cleanup.feeds_by_election(state, vip_id, current_run, old_runs) as
+  with feeds as
+      (select r.state as state,
+              r.vip_id as vip_id,
+              r.spec_version as spec_version,
+              r.election_type as election_type,
+              r.election_date as election_date,
+              array_agg(r.id order by r.id desc) as ids
+       from results r
+       where r.complete is true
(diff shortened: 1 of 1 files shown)


sduckett:
@ddellacosta The gist of it is that the tuple `(state, vip_id, spec_version, election_type, election_date)` used to be derived from a handful of tables, but now that we're collecting those attributes in `results` during processing, the tuple is trivial, and so goes the query to find feeds representing the same data set.

tank157:
@ddellacosta I can help out there. So at the end of the updated query, see how it's using array access to bifurcate the ids into `current_runs` and `old_runs`. So the most recent 5 (because of the sorting descending) will be `current` and any more than that are old. Then the `data-processor-cleanup` project has the other bit that is run to actually delete thing here: https://github.com/votinginfoproject/data-processor-cleanup/blob/master/cleanup.sql

That essentially gets run every week, deleting from results where the id is in the `old_runs` category, and at some point a long time ago a bunch of cascade deletes were added from results to all the ancillary tables.

ddellacosta:
Okay Troy explained this to me in a bit more depth in Slack--thank you both @sduckett and @tank157 !

katrinamariehh:
LGTM 💅🏻

PR #288, opened 2018-06-22
adds check to confirm we are not trying to process a zipfile above a given size

ddellacosta:
https://www.pivotaltracker.com/story/show/157016498

Does what it says on the tin.

Note that I've already added a `data-suite/max-zipfile-size` configuration value to both staging and production environments via consul.


src/vip/data_processor.clj
```diff
@@ -24,7 +24,7 @@
    t/assert-filename
    psql/start-run
    t/download-from-s3
-   zip/assoc-file
+   #(zip/assoc-file % (config [:max-zipfile-size] 3221225472)) ; 3GB default
```
tank157: Any particular reason to go with the anon function over something like `(partial zip/assoc-file (config ...))`? It just seems like the tool I would have reached for, but curious if there was a reason to go this route instead.

dev-src/dev/core.clj
```diff
@@ -18,7 +28,7 @@
 (def pipeline
   [set-context
    psql/start-run
-   zip/assoc-file
+   #(zip/assoc-file (config [:max-zipfile-size] 3221225472)) ; 3GB
```
tank157: Is this missing the `%`?

src/vip/data_processor.clj
ddellacosta: Oh, mostly because the argument is in the first position rather than at the end--or are you saying you'd rather see the arguments swapped in the `assoc-file` function signature itself?

dev-src/dev/core.clj
ddellacosta: Yes, good catch! Fixed.

src/vip/data_processor.clj
tank157: Yeah, the signature change may have been reason enough to do it this way, I think all/most other pipeline functions take the context in the first position so it would be an outlier.

PR #289, opened 2018-06-27
Fix 3 0 csv seg fault

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/158673297)


project.clj
```diff
@@ -25,7 +25,7 @@
                  [org.clojure/java.jdbc "0.6.1"]
                  [org.postgresql/postgresql "42.0.0" :exclusions [org.slf4j/slf4j-simple
                                                                   org.slf4j/slf4j-api]]
-                 [org.xerial/sqlite-jdbc "3.21.0.1"]
+                 [org.xerial/sqlite-jdbc "3.23.1"]
```
ddellacosta: This is just an opportunistic bumping of the version now that you're bypassing it by making the field a string, do I have all of that right?

tank157: Yeah. I mean, it was just on the path of upgrading things anyway, and my tests of the fix have all been with the most up-to-date versions, so it seemed ok to keep around.

PR #290, opened 2018-06-29
Add EDN reader to `max-zipfile-size` config

ddellacosta:
This should have had the `edn` reader, that's probably why this was breaking. Leaving the coercion check as it should only add more safety, but figured this should be in there too.

Related to original PR for: https://www.pivotaltracker.com/story/show/157016498


ddellacosta:
@tank157 Not at all! Thanks 👍

PR #293, opened 2018-09-19
Ack sqs messages after processing begins

tank157:
This wraps up the work to prevent zombie feeds after 12 hours of processing.

The new version of `squishy` introduces the `{:delete-callback true/false}` option, and when set to true, calls the processing function with two parameters: the message and a callback function the processor calls when it has safely received the message. This allows the processor to, say, record the start of the processing in a db and then call the delete callback to delete the message out of SQS. In this way, feed messages should never get lost in, say, a shutdown or whatever, and we no longer will spawn zombies in 12 hours on the larger feeds.

I had to replace `clj-aws-s3` because it was too old and pulling in very old incompatible dependencies, so upgrading to `amazonica` seemed appropriate and wasn't that difficult to do (with the minor annoyance now that we have to configure the region as `US_EAST_1` for squishy and `us-east-1` for amazonica.

I've already added the new configuration to both staging and production.

[Pivotal Card](https://www.pivotaltracker.com/story/show/158603633)

### DIFF
--- a/src/vip/data_processor.clj
+++ b/src/vip/data_processor.clj
@@ -19,10 +19,23 @@
             [vip.data-processor.validation.zip :as zip])
   (:gen-class))
 
+(defn ack-sqs-message
+  "If we were configured with a delete-callback from SQS, call it, effectively
+   acking that data-processor has received and kicked off processing.
+
+   This will allow us to process very long feeds without spawning zombie
+   processes after 12 hours, the maximum time we could keep a message
+   checked out."
+  [ctx]
+  (when-let [delete-callback (:delete-callback ctx)]
+    (delete-callback))
+  ctx)
+
 (def download-pipeline
   [t/read-edn-sqs-message
    t/assert-filename
    psql/start-run
+   ack-sqs-message
    t/download-from-s3
    #(zip/assoc-file % (config [:max-zipfile-size] 3221225472)) ; 3GB default
    zip/extracted-contents])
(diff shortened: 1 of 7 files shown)


src/vip/data_processor.clj
```diff
@@ -66,23 +79,27 @@
...
+  ([message delete-callback]
+   (log/info "processing SQS message" (pr-str message))
+   (q/publish {:initial-input message
+               :status :started}
+              "processing.started")
+   (let [result (pipeline/process pipeline message delete-callback)]
+     (psql/complete-run result)
+     (log/info "New run completed:"
+               (psql/get-run result))
+     (let [exception (:exception result)
```
ddellacosta: I realize you didn't introduce this in this PR, but I'd love to see these `let` statements merged for clarity's sake, e.g.:

```clojure
(let [result (pipeline/process pipeline message delete-callback)
      _ (psql/complete-run result)
      _ (log/info "New run completed:" (psql/get-run result))
      exception (:exception result)
      ... ]
```

I also find the amount of indenting in this function to be overzealous (e.g. `(q/publish completed-message "processing.complete")` is easier to parse visually when it's on a single line IMHO), but that is certainly more subjective.

src/vip/data_processor.clj
```diff
@@ -19,10 +19,23 @@
...
 
+(defn ack-sqs-message
+  "If we were configured with a delete-callback from SQS, call it, effectively
+   acking that data-processor has received and kicked off processing.
+
+   This will allow us to process very long feeds without spawning zombie
+   processes after 12 hours, the maximum time we could keep a message
+   checked out."
+  [ctx]
+  (when-let [delete-callback (:delete-callback ctx)]
```
ddellacosta: Just noticed that you could probably shrink this up like

```clojure
[{:keys [delete-callback] :as ctx}]
(when delete-callback (delete-callback))
ctx)
```

...if you so desire.

tank157: Oh, nice suggestion. And I'll see if I can clean up that other area too

PR #294, opened 2018-09-21
add migration for foreign key constraint to clean up validations table when results are deleted

ddellacosta:
https://www.pivotaltracker.com/story/show/160106899

Tested these locally FWIW, both as migrations (at least, the up migration) and manually using `psql` client directly (both up and down migrations).

### DIFF
--- a/resources/migrations/20180921-00-validations-table-result-id-cascade-delete.up.sql
+++ b/resources/migrations/20180921-00-validations-table-result-id-cascade-delete.up.sql
@@ -1,0 +1,1 @@
+alter table validations add foreign key (results_id) references results(id) on delete cascade;
--- a/resources/migrations/20180921-00-validations-table-result-id-cascade-delete.down.sql
+++ b/resources/migrations/20180921-00-validations-table-result-id-cascade-delete.down.sql
@@ -1,0 +1,1 @@
+alter table validations drop constraint validations_results_id_fkey;


resources/migrations/20180921-00-validations-table-result-id-cascade-delete.up.sql
```diff
@@ -0,0 +1 @@
+alter table validations add foreign key (results_id) references results(id) on delete cascade;
```
tank157: Can you just reassure me that this won't fail on a production db where some rows that this foreign key gets added to won't complain if, for example, we've already deleted the row in results? I _think_ it'll be ok, but i'm always fuzzy about just when adding things like this to an imperfect db (ie a production one that we've manually deleted things sometimes) will fail.

ddellacosta: Good question; I don't actually know the answer so I'll check it out and get back to you!

ddellacosta:
@tank157 Sorry this took me so long to respond to!

So you were right to be concerned, this migration will fail if we have a row with a non-existent `results_id`:

```sql
dataprocessor-# ;
 results_id | severity | scope | identifier | error_type | error_data
------------+----------+-------+------------+------------+------------
          1 | whatevs  | yeah  |          1 | okay       | something
        100 | whatevs  | yeah  |        100 | okay       | something
(2 rows)

dataprocessor=# \d validations
                      Table "public.validations"
   Column   |          Type          | Collation | Nullable | Default
------------+------------------------+-----------+----------+---------
 results_id | integer                |           |          |
 severity   | character varying(255) |           |          |
 scope      | character varying(255) |           |          |
 identifier | bigint                 |           |          |
 error_type | character varying(255) |           |          |
 error_data | text                   |           |          |
Indexes:
    "validations_result_scope_id_idx" btree (results_id, scope, identifier)
    "validations_result_scope_type_idx" btree (results_id, scope, error_type)
    "validations_results_id_severity_idx" btree (results_id, severity)
    "validations_results_scoped_severity_idx" btree (results_id, severity) WHERE severity::text = 'critical'::text OR severity::text = 'fatal'::text

dataprocessor=# alter table validations drop constraint validations_results_id_fkey;
ERROR:  constraint "validations_results_id_fkey" of relation "validations" does not exist
dataprocessor=# alter table validations add foreign key (results_id) references results(id) on delete cascade;
ERROR:  insert or update on table "validations" violates foreign key constraint "validations_results_id_fkey"
DETAIL:  Key (results_id)=(100) is not present in table "results".
dataprocessor=#
```

What I propose is that I go ahead and merge this in after I've done a cleanup of the `validations` table again on a weekend, and that way we can make sure this does the right thing. Does that seem reasonable?

tank157:
Yup, that seems reasonable! I'm glad past experience me had some insight into the ways these things can fail with a live db!

PR #302, opened 2019-03-21
get unprocessed file bucket from SQS message

tank157:
While not strictly needed for dasher to kick off processing so long as dasher and data-processor are configured with the same bucket, this is a change that's been long overdue to get the filename AND bucket from the processing message rather than just the filename. So long that the other projects that made a similar move have probably stopped using SQS long ago.

### DIFF
--- a/src/vip/data_processor/s3.clj
+++ b/src/vip/data_processor/s3.clj
@@ -28,7 +28,7 @@
 (def tmp-path-prefix "vip-data-processor")
 
 (defn download
-  "Downloads the file named `key` from the configured S3 bucket to a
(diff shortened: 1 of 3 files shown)


src/vip/data_processor/s3.clj
```diff
@@ -30,10 +30,10 @@
 (defn download
   "Downloads the file named `key` from the configured S3 bucket to a
   temporary file and returns that path."
-  [key]
+  [key bucket]
```
sduckett: does `bucket` deserve mention in the docstring?

PR #308, opened 2019-06-24
Wabbit season

tank157:
Removes the use of rabbitmq from the application and replaces it with various AWS services (SNS mostly). A lot of refactoring and then tweaks to get it operating correctly in the staging environment. Tackled a number of configuration streamlines and version upgrades as well. Oh, and removed the not really used/useful newrelic integration as well.

[Pivotal card](https://www.pivotaltracker.com/story/show/166622758)

Pairs with branches of the same name in `Metis` and `batch-address-test-tool`.


src/vip/data_processor/queue.clj
```diff
@@ -1,44 +1,54 @@
...
-              (pr-str payload)
-              {:content-type "application/edn" :type "qa-engine.event"}))
+(defn default-sns-client
+  []
+   ;; memoize the default sns-client construction, so that
+   ;; we can avoid creating extra thread pools over time.
+   ;; see https://github.com/cognitect-labs/aws-api/issues/80
+   (memoize (sns-client (config [:aws :creds :access-key])
+                        (config [:aws :creds :secret-key])
+                        (config [:aws :region]))))
```
tie-rack: `memoize` takes a function, but if I'm reading everything right, `sns-client` will return a Client object. Does this work?

Either way, another option here, since `memoize` "should" be used for referentially transparent functions, is to use [`delay`](http://clojure.github.io/clojure//clojure.core-api.html#clojure.core/delay) and deref the result where it's needed:

```clj
(def default-sns-client
  (delay (sns-client (config [:aws :creds :access-key])
;; ...))))

(defn publish-success
  "Publish a successful feed processing message to the topic."
  ([payload]
   (publish-success @default-sns-client
                    (config [:aws :sns :success-topic-arn])
                    payload))
;; ...))))
```

src/vip/data_processor/queue.clj
```diff
@@ -1,44 +1,54 @@
...
-          (do (Thread/sleep (* attempt 1000))
-              (recur (inc attempt)))
-          (do (log/error "Connecting to RabbitMQ failed. Bailing.")
-            (throw (ex-info "Connecting to RabbitMQ failed" {:attemts attempt})))))))
-  (reset! rabbit-channel
-          (let [ch (lch/open @rabbit-connection)]
-            (le/topic ch (config [:rabbitmq :exchange]) {:durable false :auto-delete true})
-            (log/info "RabbitMQ topic set.")
-            ch)))
+(def sns-client-memo (memoize sns-client))
```
tie-rack: This doesn't appear to be used anywhere.

src/vip/data_processor/queue.clj
tank157: You know, it seemingly works with `memoize`, but it never sat right with me. Your suggestion of `delay` with `def` makes much more sense, as it solves the problem of not having the config available until runtime. I'm going to give that a try and if it works, port it over to `batch-address-test-tool` as well.

I do only want one client created, as there's an open issue on the AWS library about the clients opening threadpools and not cleaning them up, so opening a new client on each request will ultimately result in OutOfMemoryError.

tank157: @tie-rack Thanks for the suggested use of `delay`, it seems to be working like a charm and I feel much better about it than `memoize` for sure!

PR #309, opened 2019-07-17
Pipeline Separation PR

tank157:
Big refactor/upgrade branch here, around the theme of improving processing performance, in particular the 5.X series.

Included are:
* Recording locality stats timing so we can pin down some missing indexes
* Major refactor to tone down the pipeline jumping from place to place all the time, but rather
  codify the pipelines into just a few mostly concrete places. After passing through some common
  pipeline infrastructure for processing queue messages, downloading files, and then determining
  which format (csv/xml) and which spec version (3.0/5.1), we jump to a specific pipeline for each of
  the 4 possible combinations. This makes it somewhat easier to follow and reason about just which
  steps are in play.
* Lots of version upgrades to reduce the vulnerabilities, now no MAJOR vulnerabilities and just a
  handfull of MEDIUM ones (down from like 35+ when this branch started)
* ~(Coming next): For 5.X processing pipelines, dropping the compound indexes, ie `ON (results_id, X)` and replacing them with feed specific indexes added at the start of processing, ie `ON (X) WHERE results_id = 123` and dropped at the end. Preliminary tests show this to be faster, and also hopefully easier to implement than creating new tablespaces for each run.~


tank157:
@lizstarin In case you have time/boredom to get started learning `data-processor` while I'm gone, this PR makes a lot of changes I want to incorporate in the future. It no longer really does much with performance improvements in 5.1 per se, but I think the refactoring to make the pipelines more explicit could be very useful.

Pipeline processing starts in `src/vip/data_processor.clj` with a pretty small pipeline of common elements + a pipeline chooser step which determines which of the four main pipelines to continue with (3.0 CSV, 3.0 XML, 5.1 CSV, 5.1 XML). Each of those specific pipelines are then very well spelled out, rather than implemented with a bunch of functions that incrementally add bits of pipeline all over the place. Comparing this to master would be a good way to get a feel not only for what processing looks like, but how (I hope) this proposal is better. It would make writing some specs for the 5.1 pipelines easier for sure.

If you do spend some time looking into this and have burning questions while I'm out, feel free to reach out to me, but also Chris Shea can be a good resource, as he created the actual pipeline processing code in the first place. This PR changes some elements of that, but the basic concept is still the same.

PR #314, opened 2020-06-04
Fix xml filename

tank157:
* Upgrade zip4j dependency
* Make the filename generation a bit more generic so we can use it for both XML and ZIP filenames
* Rename the temp xml file from the pseudo-random name that all temp files get by default to the well structured name

### DIFF
--- a/src/vip/data_processor/output/xml_helpers.clj
+++ b/src/vip/data_processor/output/xml_helpers.clj
@@ -1,12 +1,74 @@
 (ns vip.data-processor.output.xml-helpers
-  (:require [korma.core :as korma])
-  (:import [java.nio.file Files]
+  (:require [clojure.string :as str]
+            [korma.core :as korma]
+            [vip.data-processor.db.postgres :as postgres]
+            [vip.data-processor.s3 :as s3]
+            [vip.data-processor.util :as util])
+  (:import [java.nio.file Files CopyOption StandardCopyOption]
            [java.nio.file.attribute FileAttribute]))
 
+(defn format-fips [fips]
+  (cond
+    (empty? fips) "XX"
+    (< (count fips) 3) (format "%02d" (Integer/parseInt fips))
+    (< (count fips) 5) (format "%05d" (Integer/parseInt fips))
+    :else fips))
+
+(defn format-state
+  [state]
+  (if (empty? state)
+    "YY"
+    (clojure.string/replace (clojure.string/trim state) #"\s" "-")))
+
+(defn filename* [fips state election-date]
+  (let [fips (format-fips fips)
+        state (format-state state)
+        date (util/format-date election-date)]
+    (str/join "-" ["vipfeed" fips state date])))
+
+(defn generate-file-basename
+  [{:keys [spec-version tables import-id] :as ctx}]
+  (condp = (util/version-without-patch @spec-version)
+    "3.0"
+    (let [fips (-> tables
+                   :sources
+                   korma/select
+                   first
+                   :vip_id)
+          state (-> tables
+                    :states
+                    korma/select
+                    first
+                    :name)
+          election-date (-> tables
+                            :elections
+                            korma/select
+                            first
+                            :date)]
+      (assoc ctx :output-file-basename
+             (filename* fips state election-date)))
+
+    "5.1"
+    (let [fips (postgres/find-value-for-simple-path
+                import-id "VipObject.Source.VipId")
+          state (postgres/find-value-for-simple-path
+                 import-id "VipObject.State.Name")
+          election-date (postgres/find-value-for-simple-path
+                         import-id "VipObject.Election.Date")]
+      (assoc ctx :output-file-basename
+             (filename* fips state election-date)))))
+
 (defn create-xml-file
-  [{:keys [import-id] :as ctx}]
-  (let [xml-file (Files/createTempFile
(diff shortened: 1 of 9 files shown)


src/vip/data_processor/output/xml_helpers.clj
```diff
@@ -1,12 +1,74 @@
...
+            [vip.data-processor.s3 :as s3]
+            [vip.data-processor.util :as util])
+  (:import [java.nio.file Files CopyOption StandardCopyOption]
            [java.nio.file.attribute FileAttribute]))
 
+(defn format-fips [fips]
+  (cond
+    (empty? fips) "XX"
+    (< (count fips) 3) (format "%02d" (Integer/parseInt fips))
+    (< (count fips) 5) (format "%05d" (Integer/parseInt fips))
```
alpha-dw, line 14: Contextual question: Is the check for less than 5 made to account for a previous version of FIPS codes ([is this the one?](https://en.wikipedia.org/wiki/FIPS_county_code)? I am wondering why it is not checking for less than 6, but perhaps that is due to my lack of context!

tank157: TBH I'm not sure, this was part of some code I simply moved to a new location. If I had to guess, though, I'd suspect it's to handle one of several ways in which FIPS codes for Alabama, Alaska, Arizona, Arkansas, California, Colorado and Connecticut counties, all of whom start with a leading 0. It's super easy for a human to not understand that the leading 0 is important, or for a number parser to strip it off. There's still some troubling gaps in there, as mostly that case I would expect to be exactly 4 digits, but I can't figure out a valid case where we'd only have 3 digits, aside from a county accidentally _only_ providing the county portion of their full FIPS.

alpha-dw: 🤔

tank157: This commit from the original place the code appears is interesting: https://github.com/votinginfoproject/data-processor/commit/a6a10e3abe21b5ab75bc00a298d42cdea7dd39e4

So it definitely started out as a "zero-padding" thing from what I can tell.

alpha-dw: Oh, that's an interesting fix. I wonder what the case was...

tank157: I think the comment on the commit helps clarify. When the first comparison was "less than 2", all 2 digit FIPS codes were getting 0-padded to 5 digits, but all state FIPS codes are actually 2 digits so they were getting needlessly padded.

alpha-dw: Ah yes, missed that. That makes sense. And given the context you provided regarding the various FIPS, it looks clear enough for me

src/vip/data_processor/output/xml_helpers.clj
```diff
@@ -1,12 +1,74 @@
...
+             (filename* fips state election-date)))))
+
 (defn create-xml-file
-  [{:keys [import-id] :as ctx}]
-  (let [xml-file (Files/createTempFile
-                  (str import-id) ".xml" (into-array FileAttribute []))]
+  "Creates a temp file for the output xml, then moves it to a name without
+   the randomness that temp files get automatically added to their names."
+  [{:keys [import-id output-file-basename] :as ctx}]
+  (let [tmp-xml-file (Files/createTempFile
```
lizstarin, line 65: Java doesn't let you create a temp file where you just supply the full name, huh?

test/vip/data_processor/output/xml_helpers_test.clj
```diff
@@ -1,9 +1,38 @@
...
+
+  (testing "workably formatted dates make nice filenames"
+    (is (= "vipfeed-52-FL-2015-03-27"
+           (filename* "52" "FL" "2015/03/27"))))
+
+  (testing "poorly formatted dates use a workable filename"
+    (is (= "vipfeed-12345-North-Carolina-"
+           (filename* "12345" "North Carolina" nil))))
+
+  (testing "missing the fips and/or state doesn't break the world"
```
lizstarin, line 19: 😆

src/vip/data_processor/output/xml_helpers.clj
tank157: Sadly no. I suspect it's to allow you to create many temp files with the same code over and over again, but it's annoying for sure.

src/vip/data_processor/output/xml_helpers.clj
```diff
@@ -1,12 +1,74 @@
...
-  (:import [java.nio.file Files]
+  (:require [clojure.string :as str]
+            [korma.core :as korma]
+            [vip.data-processor.db.postgres :as postgres]
+            [vip.data-processor.s3 :as s3]
+            [vip.data-processor.util :as util])
+  (:import [java.nio.file Files CopyOption StandardCopyOption]
            [java.nio.file.attribute FileAttribute]))
 
+(defn format-fips [fips]
```
lizstarin, line 10: Were there tests for these formatting functions, previously?

tank157: Not that I saw. Happy to add one though

tank157: Added some tests for the formatting functions.

PR #321, opened 2020-10-12
Vip 53 more zip bomb protection

tank157:
Instead of just checking the size of the first file, sum up all the uncompressed sizes to get the uncompressed size of the entire zip. This should close the remaining holes in the zip bomb check.


lizstarin (approved):
🤐 💣

PR #330, opened 2021-02-15
Form-encode the generated database connection string

gws:
While this technically is not URL encoding, this is also the method of decoding that some (most?) JDBC drivers (checking PostgreSQL and MySQL, but probably others) use, and so it will work for most use cases. Critically, hand-written parsers seen in JDBC driver code typically look for `/`, `&`, and `=`, so as long as these are encoded using a method that java.net.URLDecoder can decode, everything will work nicely.

# References

- [VIP-21](https://democracyworks.atlassian.net/browse/VIP-21)

### DIFF
--- a/test/vip/data_processor/db/postgres_test.clj
+++ b/test/vip/data_processor/db/postgres_test.clj
@@ -1,6 +1,18 @@
 (ns vip.data-processor.db.postgres-test
   (:require [vip.data-processor.db.postgres :refer :all]
+            [turbovote.resource-config :refer [config]]
             [clojure.test :refer :all]))
+
+(deftest url-test
+  (let [db-spec-1 {:host "host"
+                   :port "5678"
+                   :user "us&r"
+                   :password "p&ssw@rd"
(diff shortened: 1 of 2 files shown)


tank157 (approved):
👾

test/vip/data_processor/db/postgres_test.clj
```diff
@@ -1,7 +1,19 @@
...
+(deftest url-test
+  (let [db-spec-1 {:host "host"
+                   :port "5678"
+                   :user "us&r"
+                   :password "p&ssw@rd"
+                   :database "database"}]
+    (with-redefs [config (constantly db-spec-1)]
+      (is (= "jdbc:postgresql://host:5678/database?user=us%26r&password=p%26ssw%40rd"
+             (url))
+          "Percent encoding is done for username and password values"))))
```
jimkoglerdw, line 15: test case answered my only question!  👍

jimkoglerdw (approved):

PR #331, opened 2021-02-16
Use AWS_REGION instead of VIP_DP_AWS_REGION

gws:
This brings us more in line with AWS environment variables and allows us to only have to specify the AWS region once in the deployment configuration.

# References

- [VIP-20](https://democracyworks.atlassian.net/browse/VIP-20)

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ project. Create that file, copy the following into it and provide
 values for each environment variable.
 
 ```
+AWS_REGION=
 DB_PORT_5432_TCP_ADDR=
 DB_ENV_POSTGRES_USER=
 DB_ENV_POSTGRES_PASSWORD=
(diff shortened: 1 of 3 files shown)


tank157 (approved):
LGTM, though it makes me wonder if we should normalize the other env var names too, now this one sticks out a bit. Maybe for another time, or just something to not repeat in the new platform.

gws:
@tank157 I considered that, and since you mention it I think I should just go ahead and do it. For the future, though, ideally we'd be sourcing credentials from the environment rather than passing them explicitly so we could make use of EC2 instance profiles, assumed roles via `aws-vault`, etc. without any code changes, so that's the direction we'd want to go in.

tank157:
@gws If that's the case, then we could skip the interim update and just wait until we could make that better improvement instead. Ideally we'll just get data-processor into a relatively steady state where we don't need to make many changes to it for the time we're working on the replacement, so I suppose the question becomes one of whether or not any improvements we make now are to achieve that stability or are we still in a mindset of improving for the long term, when that long term does have a hopefully defined endpoint now?

jimkoglerdw (approved):

gws:
@tank157 Excellent! I'm not aware of this improving the stability of `data-processor`, so I may just close this one out later today. I'll bring it up at our upcoming tech meeting.

PR #333, opened 2021-06-02
Vip 370 locality stats logging

tank157:
Moved a loop over all the localities that was inside a db function out into Clojure so we could log progress.

Slightly less efficient to do it this way, might want to do some staging testing with a WI for MI feed with a lot of localities and compare before and after applying this. I mentioned in the commit message that maybe we plan on some kind of batching operation instead, where we send groups of ids, as that will cut down on the db connection penalty a bit while still providing some insights.

[Jira Card](https://democracyworks.atlassian.net/browse/VIP-370)

### DIFF
--- a/resources/migrations/20210521-01-v5-street-segment-validations.up.sql
+++ b/resources/migrations/20210521-01-v5-street-segment-validations.up.sql
@@ -1,0 +1,19 @@
+create table v5_street_segment_validations (
+    results_id bigint NOT NULL,
+    path ltree,
+    severity character varying(255) COLLATE pg_catalog."default",
+    scope character varying(255) COLLATE pg_catalog."default",
+    error_type character varying(255) COLLATE pg_catalog."default",
+    error_data text COLLATE pg_catalog."default",
+    parent_element_id text COLLATE pg_catalog."default",
+    CONSTRAINT xml_tree_validations_results_id_fkey FOREIGN KEY (results_id)
+        REFERENCES public.results (id) MATCH SIMPLE
+        ON UPDATE NO ACTION
+        ON DELETE CASCADE
+)
+WITH (
+    OIDS = FALSE
+)
+TABLESPACE pg_default;
+
+create index v5_street_segment_paths on v5_street_segment_validations using gist (path);
--- a/resources/migrations/20210521-01-v5-street-segment-validations.up.sql
+++ b/resources/migrations/20210521-01-v5-street-segment-validations.up.sql
@@ -1,3 +1,5 @@
+create extension btree_gist;
+
 create table v5_street_segment_validations (
     results_id bigint NOT NULL,
     path ltree,
@@ -16,4 +18,4 @@ WITH (
 )
 TABLESPACE pg_default;
 
-create index v5_street_segment_paths on v5_street_segment_validations using gist (path);
+create index v5_street_segment_paths on v5_street_segment_validations using gist (results_id, path);
--- a/resources/migrations/20210521-01-v5-street-segment-validations.up.sql
+++ b/resources/migrations/20210521-01-v5-street-segment-validations.up.sql
@@ -1,4 +1,4 @@
-create extension btree_gist;
+create extension if not exists btree_gist;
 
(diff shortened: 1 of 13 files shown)


tank157:
screenshot_991
Example of the new logging

gws (approved):
I am not going to have a valuable review for all of the SQL since I don't understand the data model that well, but based on my understanding of what you were doing, the Clojure code looks how I'd expect.

resources/migrations/20210521-01-v5-street-segment-validations.up.sql
```diff
@@ -0,0 +1,19 @@
+create table v5_street_segment_validations (
+    results_id bigint NOT NULL,
```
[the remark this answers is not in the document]
tank157: Hmmmm, good points, let me look into this. It will have multiple feeds in the table at the same time potentially, so _sometimes_ it might be useful to have an index.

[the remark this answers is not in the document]
gws: For sure, and if it's going to be heavy on the writes and light on the reads, an index could slow down the process since it'll have to be maintained (and then not subsequently used very often), so there might be really good reasons _not to_ create one.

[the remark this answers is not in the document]
tank157: It's "write once", read roughly as many localities as there are.

I attempted to add the results_id to the existing GiST index, and that makes sense as the query uses a compound where clause that starts with the results_id and then the path. I'm testing this out locally now. It does require the installation of an extension in the DB, but I went ahead and tested this out in staging and it installs with the `CREATE EXTENSION btree_gist` just fine, so I'm assuming I can just add that to a migration and adjust the index to include results_id. Hopefully, being a Generalized Search Tree, and having results_id being the major index column, and only at most 3 feeds processing at once, the tree will just have a root with up to 3 children, one for each feed, and then the paths branch out below those nodes. Unless I'm totally off in how it might be implemented (very probably!)

resources/migrations/20210521-03-adjust-locality-stats.up.sql
```diff
@@ -0,0 +1,205 @@
+drop function if exists v5_dashboard.locality_stats (integer, text);
```
lizstarin, line 1: Okay, I can tell by the line count that you didn't write two identical functions for the up and down migrations, but can you help me understand the difference here?

src/vip/data_processor/db/tree_statistics.clj
```diff
@@ -70,21 +71,70 @@
...
 
 (defn stats
-  [{:keys [import-id]}]
+  [import-id]
   (->> (korma/exec-raw [(error-query) [import-id import-id]] :results)
        (map stats-map)
        (reduce merge)))
 
+(defn prepare-temp-data-for-locality-stats
+  "Adds steet segment validation data to processing table for faster
```
lizstarin, line 80: tiny typo :)

resources/migrations/20210521-03-adjust-locality-stats.up.sql
tank157: Good question! Yeah, the change is fairly minor, instead of using the temp table that the previous all database function based solution uses, this change switches to use the shared `v5_street_segment_validations` table that is now a long term table we add/remove data from. I couldn't seem to make it work with temp tables created in one function and then referred to in later function calls when they weren't all enclosed in the same transaction. So the main change is around line 111 in the up migration where it now uses `public.v5_street_segment_validations` in the select query, rather than the temp table that was specific to this feed run.

src/vip/data_processor/db/tree_statistics.clj
tank157: You mean `steet` isn't the hip new slang word for `street`? Thanks for the catch! :)

gws (approved):
Sorry for getting to this so late, I had some things run long! I've used `btree_gist` (it's in production in `scheduled-message-works` in fact) and it works well.

PR #334, opened 2021-06-23
Stop feed

lizstarin:
Creates mechanism for stopping a feed run. Goes with [a PR to `Metis`. ](https://github.com/votinginfoproject/Metis/pull/511)


resources/migrations/20210527-01-add-feed-stop-flag-to-results.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE results ADD COLUMN stop_requested TEXT;
```
tank157, line 1: I _think_ maybe the TEXT column will default to null, but it never hurts to explicitly set it to default to null here, especially as later on any non-nil value will trigger a stop.

src/vip/data_processor/db/postgres.clj
```diff
@@ -343,6 +342,18 @@
   (korma/select results
                 (korma/where {:id (:import-id ctx)})))
 
+(defn get-run-field [ctx field]
```
tank157, line 345: The use of `run` as a name here is giving me some pause. I know we colloquially refer to them as feed runs when discussing things, but the code not as much. So it sort of gives me pause as to thinking about what does run mean here, and what does it have to do with the results table. Maybe just adding a docstring explaining it might help, because I'm not against using the term, it just is a tad bit confusing, and may be confusing in the future when revisiting.

src/vip/data_processor/pipeline.clj
```diff
@@ -18,6 +18,13 @@
                  :thrown-by processing-fn
                  :exception e))))
 
+(defn check-stop-flag
+  [ctx]
+  (log/info "Checking stop request flag")
```
tank157, line 23: Small suggestion, maybe use a logging function like `log/debug` instead. A generalish rule of thumb around logging is that things that repeat without extra info (like, "I'm in this function!") are good to set at a debug level so production systems can filter them out. It's not always a hard/fast rule of thumb though, the locality stats logging I've been working on is also a repeating thing, but I do include info that helps someone looking at the logs make some determinations (how many localities are left to run, locality ids, etc).

That said, it would be nice to see in the logs that a stop was requested, maybe make the true branch of the if something like:
```
(do (log/info "DB stop requested!")
       (assoc ctx :stop "Stop requested"))
```

src/vip/data_processor/pipeline.clj
```diff
@@ -27,7 +34,10 @@
     (let [[next-step & rest-pipeline] (:pipeline ctx)]
       (if next-step
         (let [ctx-with-rest-pipeline (assoc ctx :pipeline rest-pipeline)
-              next-ctx (try-processing-fn next-step ctx-with-rest-pipeline)]
+              check-stop-flag-fn (:check-stop-flag-fn ctx)
+              next-ctx (cond->> ctx-with-rest-pipeline
```
tank157, line 38: Interesting use of `cond->>` here! It does make it a bit hard to read, but honestly I'm not seeing another way that significantly improves it off the top of my head. Could break out the whole section into a dedicated function, but you're still going to have to call the `try-processing-fn` and `check-stop-flag-fn` if it exists, and pass the context through them and back out, so seems ok to me.

test/vip/data_processor/pipeline_test.clj
```diff
@@ -21,28 +21,45 @@
...
+          result (run-pipeline ctx)]
+      (is (:stop result))
+      (is (= 1 (:input result)))
+      (is (not (empty? (:pipeline result))))
+      (is (not (contains? result :exception))))))
 
 (deftest process-test
   (testing "Builds the context and executes the pipeline"
-    (let [intial-input {:input 0}
+    (let [intial-input {:input 0 :check-stop-flag-fn #(identity %)}
```
tank157, line 62: Good use of identity! I think you can simplify this to just `:check-stop-flag-fn identity`, you don't need the inline function wrapped around it.

tank157 (approved):
Looks like a great start! I have some suggestions, but overall it's looking good!

🍨

src/vip/data_processor/db/postgres.clj
lizstarin: I can add a docstring; this is following the name of the function above, `get-run`.

tank157: docstring would be awesome!

tank157 (approved):
Looks great!

🛑 🥘  (Sorta my attempt at "stop feed" in emoji)

PR #343, opened 2021-07-20
Stop feed

lizstarin:


gws (approved):
@lizstarin Sorry for the late review! This LGTM. I should have GitHub email me when I let review requests go stale - please always feel free to ping if my review is blocking something.

PR #348, opened 2021-08-27
Avoid XXE in clojure.data.xml

gws:
XXE (an **X**ML e**x**ternal **e**ntity attack) is possible in `clojure.data.xml` versions 0.0.8 and below. We are unable to easily upgrade to a more recent version (and no more recent "stable" version exists).

We are introducing a wrapper namespace wrapping functions in `clojure.data.xml` with safer XML parsing functions that apply XXE protection by default, with the goal of ultimately removing the wrapper namespace when the application is updated.

# References

- [VIP-386](https://democracyworks.atlassian.net/browse/VIP-386)

### DIFF
--- a/src/vip/data_processor/validation/xml.clj
+++ b/src/vip/data_processor/validation/xml.clj
@@ -1,12 +1,12 @@
 (ns vip.data-processor.validation.xml
-  (:require [clojure.data.xml :as xml]
-            [clojure.walk :refer [stringify-keys]]
+  (:require [clojure.walk :refer [stringify-keys]]
             [korma.core :as korma]
             [vip.data-processor.db.postgres :as postgres]
             [vip.data-processor.db.sqlite :as sqlite]
             [vip.data-processor.util :as util]
             [vip.data-processor.validation.data-spec :as data-spec]
-            [vip.data-processor.errors :as errors])
+            [vip.data-processor.errors :as errors]
+            [vip.data-processor.xml :as xml])
   (:import [java.nio.file Files StandardCopyOption]))
 
 (def address-elements
--- a/src/vip/data_processor/validation/xml/v5.clj
+++ b/src/vip/data_processor/validation/xml/v5.clj
@@ -1,10 +1,10 @@
 (ns vip.data-processor.validation.xml.v5
-  (:require [clojure.data.xml :as xml]
-            [clojure.string :as str]
+  (:require [clojure.string :as str]
             [vip.data-processor.db.postgres :as postgres]
             [vip.data-processor.util :as util]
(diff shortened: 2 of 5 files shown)


alpha-dw:
LGTM! Please see notes below for a potential next step.

*Note 1: While performing local runs on 3.0, 5.1.1, 5.1.2, 5.1, and 5.2 XML test files, I noticed a less-than-desired behavior with 3.0 potentially-containing-xxe XML feeds (details below).
The expected behavior occurs for all v5 feeds: stopping the current run with a descriptive error message and prevention of the creation of an output file.
However, for v3 feeds, such desired behavior is lacking, and in fact the run appears superficially "normal" instead of appearing erroneous.
This behavior pre-exists this XXE issue, so perhaps we can address this v3 issue in another ticket. I can create it if we choose to do so. Please let me know your thoughts.

v3 XXE File
- DOES NOT STOP RUN, NO DESCRIPTIVE ERROR MESSAGE
- Leads to creation of output xml file of the following structure:
    Filename: vipfeed-XX-YY-.xml
    which contains
    ```xml
    <?xml version='1.0' encoding='UTF-8'?>
    <vip_object xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance' xsi:noNamespaceSchemaLocation='http://election-info-standard.googlecode.com/files/election_spec_v3.0.xsd' schemaVersion='3.0'></vip_object>
    ```

v5 XXE File
- Stops run with the following logs
- Does not create output xml file

> 11:17:20.436 INFO  [vip.data-processor.pipeline] (main) {:import-id 18}
> 11:17:20.461 ERROR [vip.data-processor.pipeline] (main) Stopping run of 18 due to: Exception caught
> 11:17:20.497 ERROR [vip.data-processor.pipeline] (main) com.ctc.wstx.exc.WstxParsingException: Undeclared general entity "xxe"


*Note 2: The processing of a potential XXE-containing file leads to similar behavior as dereferencing an undeclared entity (i.e. only &myref;).

gws:
>However, for v3 feeds, such desired behavior is lacking, and in fact the run appears superficially "normal" instead of appearing erroneous.
>This behavior pre-exists this XXE issue, so perhaps we can address this v3 issue in another ticket. I can create it if we choose to do so. Please let me know your thoughts.

Good find! Yes, I think this should be addressed separately because it sounds like we'd be addressing the fact that XML parser-level errors apparently don't stop v3 runs, as opposed to anything specific to the XXE case which this is designed to resolve. Thank you for doing the deeper dive!

Is there anything you need me to change or fix in order to get a 👍🏻?

alpha-dw:
Nope, you're good to go!

alpha-dw (approved):
👍

PR #353, opened 2022-05-24
VIP-534: GIS pilot branch is ready for validation writing

lizstarin:
https://democracyworks.atlassian.net/browse/VIP-534


resources/specs/vip_spec_v6.0.xsd
```diff
@@ -0,0 +1,732 @@
...
+
+        <xs:element name="State" type="State" />
+
+        <xs:element name="StreetSegment" type="StreetSegment" />
+
+      </xs:choice>
+      <xs:attribute name="schemaVersion" type="xs:decimal" use="required" fixed="5.2" />
+    </xs:complexType>
+  </xs:element>
+</xs:schema>
```
alice-dw, line 732: anything to check for in this file or has it been tested/looked at already?

dev-src/dev/core.clj
```diff
@@ -90,8 +90,8 @@
   (psql/initialize)
   (let [file (java.nio.file.Paths/get filename (into-array [""]))
         initial-context {:file file
-                         :post-process-street-segments? false
-                         :keep-feed-on-complete? true
+                         :post-process-street-segments? true
+                         :keep-feed-on-complete? false
```
alice-dw, line 94: hm why the switch?

src/vip/data_processor/pipelines/xml/v6.clj
```diff
@@ -0,0 +1,125 @@
+(ns vip.data-processor.pipelines.xml.v6
+  (:require [vip.data-processor.cleanup]
+            [vip.data-processor.db.postgres]
+            [vip.data-processor.errors]
+            [vip.data-processor.errors.process]
+            [vip.data-processor.s3]
+            [vip.data-processor.output.xml-helpers]
+            [vip.data-processor.validation.v5.ballot-measure-contest]
```
alice-dw, line 8: is it expected that these are all v5s and not v6? (i feel like yes but just wanna make sure)

alice-dw (commented):
honestly don't know what's going on enough to have any valuable input, but left a few q's so maybe i can inch closer to knowing what's happening here

resources/specs/vip_spec_v6.0.xsd
lizstarin: Interesting question, because this is actually _not_ the updated spec. It's just a clone of 5.2, serving as a placeholder for now. We don't need it yet.

src/vip/data_processor/pipelines/xml/v6.clj
lizstarin: Yes. We are cloning all the validations in the v5 pipeline, and we will add additional validations for v6.

danny-pease (commented):
Looks good! Just a couple questions, none of which necessarily block merging

src/vip/data_processor/util.clj
```diff
@@ -55,4 +55,22 @@
...
+      .toFile
+      .getName))
+
+(defn file-extension
+  "Given a file path, return the lowercased, trimmed extension (ie txt, csv, etc)"
+  [file-path]
+  (-> file-path
+      file-name
+      (str/split #"\.")
+      last
```
dwxander, line 74: do we only receive txt, csv, xml, zip files? Thinking of weird cases like .tar.gz and other possible two-dot extensions

dwxander: I think validation/transforms.clj answered this for me

dwxander (approved):
I also don't feel like I have enough insight into dataprocessor to say "this looks great!" but I don't see anything that jumps out as bad or unclear

dev-src/dev/core.clj
lizstarin: I actually don't know!

src/vip/data_processor/util.clj
lizstarin: Yeah, Franklin and I had a conversation about this because there was a test feed that had something weird like xxxx.shp.xml attached (in addition to the regular xml) and it broke the pipeline—I think where we landed is that the states should reasonably be expected to adhere to a specified folder structure and contents. We do ignore everything that isn't a type we want anyway.

PR #358, opened 2022-06-01
Clone 5.X XML pipeline to start 6.0 pipeline

danny-pease:
Adds support for processing XML files that have
the 6.0 spec version. Currently just a clone
of the 5.X XML pipeline, but can add new 
validations easily.


lizstarin:
Are you literally thinking just start with the pipeline clone, and add the other setup commits to the second PR?

danny-pease:
yup! 

> add the other setup commits to the second PR?

with the important caveat that the second PR is opened only after the first is merged

lizstarin:
Is that first commit actually a squashed commit?

danny-pease:
(if it matters: squashed via cherry-pick, not by rebasing)

lizstarin:
It's confusing that it has the same commit message as one of the original commits, though.
I do think that keeping your second commit here separate will make it easier to add that code back.

danny-pease:
> It's confusing that it has the same commit message as one of the original commits, though.

good point! i was kinda rushing and forgot to update the message

> I do think that keeping your second commit here separate will make it easier to add that code back.

aggreed!

danny-pease:
> Otherwise, looks fine to me. I'm in the habit of manually running a file just to make sure we haven't introduced some pipeline-breaking step that isn't covered in the tests.

yeah absolutely. i also didnt run any tests, this was mostly an exercise in commit wrangling

lizstarin:
should I/you go ahead and merge this?

src/vip/data_processor/pipelines/xml/v6.clj
```diff
@@ -0,0 +1,125 @@
...
+            [vip.data-processor.validation.v5.internationalized-text]
+            [vip.data-processor.validation.v5.district-type]
+            [vip.data-processor.validation.v5.office]
+            [vip.data-processor.validation.v5.party]
+            [vip.data-processor.validation.v5.party-selection]
+            [vip.data-processor.validation.v5.ordered-contest]
+            [vip.data-processor.validation.v5.street-segment]
+            [vip.data-processor.validation.v5.booleans]
+            [vip.data-processor.validation.v5.polling-location]
+            [vip.data-processor.validation.v6.external-file]
```
lizstarin, line 32: delete this line—references a namespace that's no longer there

src/vip/data_processor/pipelines/xml/v6.clj
```diff
@@ -0,0 +1,125 @@
...
+    vip.data-processor.db.postgres/store-spec-version
+    vip.data-processor.db.postgres/store-public-id
+    vip.data-processor.db.postgres/store-election-id
+    vip.data-processor.db.postgres/populate-locality-table
+    vip.data-processor.db.postgres/populate-i18n-table
+    vip.data-processor.db.postgres/populate-sources-table
+    vip.data-processor.db.postgres/populate-elections-table]
+
+   ;;add validations for the types introduced with 6.0 here
+   [vip.data-processor.validation.v6.external-file/validate-no-missing-file-names]
```
lizstarin, line 53: also this one

lizstarin (changes requested):
Couple of stray references to a now-missing namespace to get rid of!

danny-pease:
it's a friday so let's not! lets touch base on monday monring to talk about this merge

danny-pease:
> Couple of stray references to a now-missing namespace to get rid of!

nice catch!

lizstarin (commented):
ready to merge?

lizstarin:
How do we set this up so that the commits don't squash?

danny-pease:
> ready to merge?

like i said it's friday so let's not merge this and touch base on monday morning 

> How do we set this up so that the commits don't squash?

i say we squash it anyway! i _think_ we can checkout to that commit (or cherry-pick it) even the squash, and if i'm wrong about that it's a small change to reimplement anyway which makes it a good learning opportunity for me (or anyone really) to do "from scratch"

lizstarin:
> > ready to merge?
> 
> like i said it's friday so let's not merge this and touch base on monday morning
> 
> > How do we set this up so that the commits don't squash?
> 
> i say we squash it anyway! i _think_ we can checkout to that commit (or cherry-pick it) even after the squash, and if i'm wrong about that it's a small change to reimplement anyway which makes it a good learning opportunity for me (or anyone really) to do "from scratch"

In that case I may also be about to have a learning experience.

dev-src/dev/core.clj
```diff
@@ -90,8 +90,8 @@
   (psql/initialize)
   (let [file (java.nio.file.Paths/get filename (into-array [""]))
         initial-context {:file file
-                         :post-process-street-segments? false
-                         :keep-feed-on-complete? true
+                         :post-process-street-segments? true
+                         :keep-feed-on-complete? false
```
alice-dw, line 94: this is something we should look out for when we qa, or should we revert back since we don't know why this change exists?

alice-dw (approved):

dev-src/dev/core.clj
lizstarin: I originally wasn't sure about this, but the docstring (expand above) provides a lot more context. We should keep it as-is. This is the pipeline that runs locally with the command `lein run [xml filename]`, and these settings will make a feed run more quickly in the dev context.

src/vip/data_processor/pipelines/xml/v6.clj
```diff
@@ -0,0 +1,123 @@
...
+    vip.data-processor.db.postgres/store-public-id
+    vip.data-processor.db.postgres/store-election-id
+    vip.data-processor.db.postgres/populate-locality-table
+    vip.data-processor.db.postgres/populate-i18n-table
+    vip.data-processor.db.postgres/populate-sources-table
+    vip.data-processor.db.postgres/populate-elections-table]
+
+   ;;add validations for the types introduced with 6.0 here
+
+   #_[vip.data-processor.validation.v5.ballot-measure-contest/validate-ballot-measure-types
```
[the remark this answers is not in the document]
lizstarin: Yes, because the v6 pipeline will ultimately include these validations. However, there's no need to run them all while writing/debugging new ones.

dev-src/dev/core.clj
danny-pease: I wound up looking into this a little more, totally forgot to circle back to the PR - I don't think it's such a big deal because these bools define how things should behave for developing locally

But bigger picture, if something changes in the dev setup i think we can request that they be omitted from a PR if we'd like to (unless of course it's something we all would benefit from)

danny-pease (commented):
Github won't let me approve my own PR (definitely for the best), so I'm removing myself as an assignee. Fwiw lgtm!

dwxander (approved):
as far as I can tell, looks good

PR #362, opened 2022-06-15
VIP-537: Set up external-geospatial-feature validation

alice-dw:


danny-pease:
Status: Liz and I are futzing with the tickets/scopes for the GIS work and this PR may need to stretch to accommodate what we learn from that process. We need to nail this down before merging/reviewing, will follow up wiht an update soon. Thanks!

danny-pease:
Update: _mostly_ sure that this review doesn't need to change as a result of ticket rehashing. Will update again tomorrow, by when I will know for sure.

test/vip/data_processor/validation/v6/external_geospatial_feature_test.clj
```diff
@@ -0,0 +1,123 @@
...
+                        v6.external-geospatial-feature/validate-no-missing-feature-identifiers]}
+        out-ctx (pipeline/run-pipeline ctx)
+        errors (all-errors errors-chan)]
+    (are [path]
+         (assert-no-problems errors
+                             {:severity :errors
+                              :scope :external-geospatial-features
+                              :identifier path})
+      "VipObject.0.ExternalGeospatialFeature.0.ExternalFileId")
+    (are [path]
```
dwxander, line 29: General q for validations: is there a difference between doing multiple `(are` statements and putting multiple paths in one statement? I noticed both in existing tests.

test-resources/xml/v6-external-geospatial-feature.xml
```diff
@@ -0,0 +1,70 @@
...
+  <!-- 0. all fields populated -->
+  <ExternalGeospatialFeature>
+    <ExternalFileId>ef1</ExternalFileId>
+    <FileFormat>shp</FileFormat>
+    <FeatureIdentifier>
+        <Index>3</Index>
+    </FeatureIdentifier>
+  </ExternalGeospatialFeature>
+
+  <!-- 1. missing ExternalFileId -->
```
dwxander, line 13: putting the indices in is such a good idea

test/vip/data_processor/validation/v6/external_geospatial_feature_test.clj
```diff
@@ -0,0 +1,123 @@
...
+                           {:severity :errors
+                            :scope :external-geospatial-features
+                            :identifier "VipObject.0.ExternalGeospatialFeature.5.FeatureIdentifier"
+                            :error-type :missing}))
+      (is (contains-error? errors
+                           {:severity :errors
+                            :scope :external-geospatial-features
+                            :identifier "VipObject.0.ExternalGeospatialFeature.6.FeatureIdentifier"
+                            :error-type :missing}))
+      (is (contains-error? errors
```
dwxander, line 114: Does the empty FeatureIdentifier index create a missing path that this test would expect?

danny-pease (approved):
lgtm!

test-resources/xml/v6_sample_feed.xml
```diff
@@ -67,6 +67,14 @@
     </Checksum>
   </ExternalFile>
 
+  <ExternalGeospatialFeature>
```
lizstarin, line 70: This element will need its own id attribute as well! i.e. `<ExternalGeospatialFeature id="exgf01">`

test-resources/xml/v6_sample_feed.xml
```diff
@@ -67,6 +67,14 @@
...
+    <ExternalFileId>ef1</ExternalFileId>
+    <FileFormat>shp</FileFormat>
+    <FeatureIdentifier>
+        <Index>3</Index>
+    </FeatureIdentifier>
+  </ExternalGeospatialFeature>
+  
   <SpatialBoundary>
```
lizstarin, line 78: Actually, this one nested inside `SpatialBoundary` is what you want; I don't think you need to repeat `ExternalGeospatialFeature` as a top-level element.

lizstarin: (and make sure to give the `SpatialBoundary` an id attribute as well)

test/vip/data_processor/validation/v6/external_geospatial_feature_test.clj
lizstarin: Yes, empty elements simply don't become entries in the ltree table, so it's the same as if the element weren't there at all.

test-resources/xml/v6-external-geospatial-feature.xml
```diff
@@ -0,0 +1,70 @@
+<?xml version="1.0"?>
+<VipObject>
+  
+  <!-- 0. all fields populated -->
```
lizstarin, line 4: I like the way you've organized this file.

test/vip/data_processor/validation/v6/external_geospatial_feature_test.clj
```diff
@@ -0,0 +1,123 @@
...
+             :pipeline [psql/start-run
+                        xml/load-xml-ltree
+                        v6.external-geospatial-feature/validate-no-missing-file-ids]}
+        out-ctx (pipeline/run-pipeline ctx)
+        errors (all-errors errors-chan)]
+    (testing "missing ExternalFileIds are flagged"
+      (is (contains-error? errors
+                           {:severity :errors
+                            :scope :external-geospatial-features
+                            :identifier "VipObject.0.ExternalGeospatialFeature.1.ExternalFileId"
```
lizstarin, line 55: To continue the above conversation about nesting `ExternalGeospatialFeature`, I believe it's _always_ nested inside `SpatialBoundary`, in which case you should do the same in your "bad" XML as well.

src/vip/data_processor/validation/v6/external_geospatial_feature.clj
```diff
@@ -0,0 +1,39 @@
...
+   :errors :external-geospatial-features :missing :missing-file-id
+    "SELECT xtv.path
+    FROM (SELECT DISTINCT subltree(path, 0, 4) || 'ExternalFileId' AS path
+          FROM xml_tree_values WHERE results_id = ?
+          AND subltree(simple_path, 0, 2) = 'VipObject.ExternalGeospatialFeature') xtv
+    LEFT JOIN (SELECT path FROM xml_tree_values WHERE results_id = ?) xtv2
+    ON xtv.path = subltree(xtv2.path, 0, 5)
+    WHERE xtv2.path IS NULL"
+   util/two-import-ids))
+
```
lizstarin, line 16: You'd expect to see `ExternalGeospatialFeature` inside a `SpatialBoundary`, I believe, which would change these paths and indices.

lizstarin (changes requested):
Sorry this is late! Don't know how I missed it. 
Comments mostly around the nesting of `ExternalGeospatialFeature` inside `SpatialBoundary`.

src/vip/data_processor/validation/v6/external_geospatial_feature.clj
alice-dw: fixed!

test/vip/data_processor/validation/v6/external_geospatial_feature_test.clj
alice-dw: i think this may be one? https://github.com/votinginfoproject/data-processor/blob/staging/test/vip/data_processor/validation/v5/candidate_test.clj#L21-L30

lizstarin: Ah. So the trick is that `are` is already a shortcut for multiple `is` statements:https://clojure.github.io/clojure/clojure.test-api.html#clojure.test/are

So if the assertions are the same, then using multiple `are` statements is just an organizational choice, afaik.

lizstarin: (and if it's just one path, you could use `is`)

alice-dw: i see. is it preferable then to combine multiple paths into one statement? if i'm using `are` already anyways

lizstarin: I think that depends entirely on what the situation demands in terms of code conciseness vs. readability.

alice-dw:
@lizstarin ready for re-review!

test-resources/xml/v6-external-geospatial-feature.xml
lizstarin: I would add `id` attributes to `ExternalGeospatialFeature` as well.

lizstarin (changes requested):
More `id` attributes.

lizstarin (approved):
🌐

PR #365, opened 2022-06-20
Vip 611 simple checksum validation

dwxander:


test/vip/data_processor/validation/v6/checksum_test.clj
```diff
@@ -0,0 +1,96 @@
...
+                        xml/load-xml-ltree
+                        v6.checksum/validate-no-missing-algorithm]}
+        out-ctx (pipeline/run-pipeline ctx)
+        errors (all-errors errors-chan)]
+    (are [path]
+        (assert-no-problems errors
+                            {:severity :errors
+                             :scope :checksum
+                             :identifier path})
+      "VipObject.0.Checksum.0.Algorithm")))
```
alice-dw, line 26: is the second 0 here the right index? im still figuring out how exactly this indexing works but i feel like it might be 4 or something since `ExternalFile` is 3...tho i'm not sure since that seems like we need to update the index each time we add something new to the sample feed, which feels wrong

test-resources/xml/v6-checksum-without-algorithm.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<VipObject>
```
alice-dw, line 2: nit: maybe a newline after the opening `VipObject` tag to be consistent with `v6-external-file-without-uri.xml`? or in general might be good to keep these consistent (looked at other test xmls and it's all over the place so maybe nvm)

alice-dw (commented):
ok some comments, 90% of this stems from my not knowing what's going on

project.clj
```diff
@@ -32,7 +32,7 @@
                  [org.postgresql/postgresql "42.2.14"
                   :exclusions [org.slf4j/slf4j-simple
                                org.slf4j/slf4j-api]]
-                 [org.xerial/sqlite-jdbc "3.28.0"]
+                 [org.xerial/sqlite-jdbc "3.34.0"]
```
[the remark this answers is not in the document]
dwxander: When I ran lein test, it just updated versions automatically—I don't think it's at all necessary for the merge, but at some point I remember discussing wanting to leave these updates in for PRs

test/vip/data_processor/validation/v6/checksum_test.clj
```diff
@@ -0,0 +1,96 @@
...
+
+(use-fixtures :once setup-postgres)
+
+(deftest ^:postgres validate-no-missing-algorithm-success-test
+  (let [errors-chan (a/chan 100)
+        ctx {:errors-chan errors-chan
+             :xml-source-file-path (xml-input "v6_sample_feed.xml")
+             :pipeline [psql/start-run
+                        xml/load-xml-ltree
+                        v6.checksum/validate-no-missing-algorithm]}
```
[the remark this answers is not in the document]
dwxander: Yeah I was wondering about this too and definitely open to adding! I looked at the external file ur test and tried to replicate its structure which is how I ended up here, but still definitely unsure of best validation practices. I also wonder if we can skip because whenever we actually test the checksum we'll need both values anyway.

test/vip/data_processor/validation/v6/checksum_test.clj
dwxander: I thought this was supposed to reference 0th VipObject -> 0th Checksum -> Algorithm? But could be totally wrong.

dwxander: I think I was wrong! Redoing this

test/vip/data_processor/validation/v6/checksum_test.clj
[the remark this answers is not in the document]
dwxander: Consolidated these

test/vip/data_processor/validation/v6/checksum_test.clj
```diff
@@ -0,0 +1,81 @@
...
+                        v6.checksum/validate-no-missing-algorithm
+                        v6.checksum/validate-no-missing-value]}
+        out-ctx (pipeline/run-pipeline ctx)
+        errors (all-errors errors-chan)]
+    (are [path]
+        (assert-no-problems errors
+                            {:severity :errors
+                             :scope :checksum
+                             :identifier path})
+      "VipObject.0.Checksum.4.Algorithm")))
```
alice-dw, line 27: consolidated test looks nice n clean! i think we might also want to `assert-no-problems` on `VipObject.0.Checksum.4.Value` here too?

dwxander: oh shoot you're so right

danny-pease (approved):
Looks good! As with the other review, sorry I don't have any more granular feedback, I'm still getting up to speed on clojure/this repo.

alice-dw (approved):
💯

lizstarin:
Looking good. The only change it needs (and I realize this is not explicit in the tutorial) is that the `Checksum`s need id attributes, e.g. `<Source id="cs001">` or something like that.

dwxander:
> Looking good. The only change it needs (and I realize this is not explicit in the tutorial) is that the `Checksum`s need id attributes, e.g. `<Source id="cs001">` or something like that.

I did not add id attributes originally because the documentation doesn't show checksum elements with them (see attached) - happy to add them here, and we should update the docs if they're needed!
image

dwxander:
@lizstarin bumping this—do added ids look ok, and should I add `id` field to documentation?

lizstarin (approved):
👯

PR #366, opened 2022-06-22
VIP-536: Add spatial boundary validation

alice-dw:

PR #368, opened 2022-06-27
VIP-536: Spatial Boundary patch

alice-dw:
fixing https://github.com/votinginfoproject/data-processor/pull/366, tested with `lein test` locally which passed 😅


alice-dw:
besides having `lein test` passing, any other local checks we should be doing before pushing/merging to make sure everything is in a good state?

PR #372, opened 2022-07-20
VIP 631

lizstarin:
Adds XML and CSV spec files and related resources. 

I've sort of punted on the question of whether to clone everything into separate 6.0 folders: I've made a new `6-0` folder for the CSV test resources that are referenced by the CSV spec (`v6_0.clj`). The folder only contains the new/modified types, because presumably we won't need to duplicate all the tests.


alice-dw (approved):

src/vip/data_processor/validation/data_spec/v6_0.clj
```diff
@@ -329,7 +335,10 @@
               {:name "precinct_split_name"}
               {:name "ballot_style_id"}
               {:name "electoral_district_ids"}
-              {:name "polling_location_ids"}]}
+              {:name "polling_location_ids"}
+              {:name "external_file_id"}
+              {:name "file_format"}
+              {:name "feature_index"}]}
```
dwxander, line 341: to clarify, this 6.0 spec file still needs to be updated with other new types, right?

dwxander (approved):

danny-pease:
Closing this - we'll revisit this another way in future

PR #373, opened 2022-09-26
VIP-744: Add notify flag to optionally suppress dashboard emails

alice-dw:


alice-dw:
https://github.com/votinginfoproject/Metis/pull/561
https://github.com/votinginfoproject/vip_admin/pull/326

alice-dw:
This change contains 3 parts:
1. vip_admin sends the `notify?` flag to data-processor on the queue
2. data-processor does its thing and then passes that flag through to its completion message
3. the completion message gets picked up by Metis and then we use the flag to suppress emails

I tested 1-2 on QA. #3 seems hard to test in QA, so for now I believe this change is safe enough to push to staging. If for some reason the flag doesn't make it to Metis, accessing a non-existent property from a js object will just return undefined, which shouldn't break anything

src/vip/data_processor.clj
```diff
@@ -30,7 +30,8 @@
          stop (:stop result)
          completed-message {:initialInput message
                             :status "complete"
-                            :publicId (:public-id result)}]
+                            :publicId (:public-id result)
+                            :notify (:notify? message)}]
```
alice-dw, line 34: just pass `notify` instead of `notify?` here so that when we receive the message in node in metis, we can access it

danny-pease (approved):
🌵

PR #374, opened 2022-10-17
Vip 556 clean feed url

dwxander:
In current datafeed.txt, the only character that matches the regex shown below is the comma in the D.C. feed name


dwxander:
@danny-pease added logging; local dataprocessor is not uploading to s3 but saving file locally using the `zip-filename` method is working as expected.

danny-pease (approved):

PR #376, opened 2023-02-17
VIP-882: Pretty print xml

alice-dw:
you can test this locally--
`lein run "path-to-input-zip"` (i used the 2022-11-08 utah output zip i found in dropbox)
and then unzip the file written to the root directory and see the pretty printed xml file.

one thing i'm unsure of--is this the right place to be "intercepting" the file writing? it works locally and i'm 98% sure it'll work  for s3 uploads because [the upload fn calls prepare-zip-file](https://github.com/votinginfoproject/data-processor/blob/main/src/vip/data_processor/s3.clj#L90), which is where we call this new formatting code. but are there other places/ways that something could end up in s3 besides the `s3.upload-to-s3` function?


src/vip/data_processor/s3.clj
```diff
@@ -45,6 +45,39 @@
   [{:keys [output-file-basename]}]
   (str output-file-basename ".zip"))
 
+(defn ppxml
+  "From https://nakkaya.com/2010/03/27/pretty-printing-xml-with-clojure/"
```
alice-dw, line 49: i literally copy-pasted this from the internet, honestly not entirely sure the ins and outs of all these methods. are we worried about that? either in terms of security or correctness

src/vip/data_processor/s3.clj
```diff
@@ -45,6 +45,39 @@
...
+  [xml-output-file]
+  (let [
+        xml-file (if (instance? File xml-output-file)
+               xml-output-file
+               (.toFile xml-output-file))
+        xml-string (slurp xml-file)
+        pretty-xml-string (ppxml xml-string)
+        ]
+    (spit xml-file pretty-xml-string)
+    xml-file))
```
alice-dw, line 76: til-slurp and spit are clojure's way of going from string -> file and back. mildly uncomfortable naming

alice-dw: also, rn it is reading the file to a string, formatting that string, and then writing back to the file. is that the best way to do this? not sure if there is a clever way we can leverage DOM or something to make it more efficient. seems to run fine locally as is but idk

dwxander: that seems fine to me??

src/vip/data_processor/s3.clj
dwxander: i feel like this raises no concerns from me, i would expect java xml transformer to be correct enough

dwxander (approved):
💅🖨❌Ⓜ️🛴

danny-pease:
> are there other places/ways that something could end up in s3 besides the s3.upload-to-s3 function?

not that i'm aware of! i took a quick look around dasher and it seems to me that the upload functionality there strictly concerns data entering VIPland via the dashboard

LGTM! let's put this on staging and let it kick around before a prod release next week

🪴

PR #378, opened 2023-03-16
VIP-919: fix BOM bug in pretty print function

alice-dw:


src/vip/data_processor/s3.clj
```diff
@@ -51,10 +51,11 @@
     zip-name))
 
 (defn pretty-print-xml
-  "From https://nakkaya.com/2010/03/27/pretty-printing-xml-with-clojure/"
+  "Adapted from https://nakkaya.com/2010/03/27/pretty-printing-xml-with-clojure/.
+   Takes xml file as input."
   [xml]
   (let [in (javax.xml.transform.stream.StreamSource.
-            (java.io.StringReader. xml))
+            (util/bom-safe-reader xml))
```
alice-dw, line 58: instead of passing a StringReader to this function, we can pass a Reader, which is returned by bom-safe-reader

src/vip/data_processor/s3.clj
```diff
@@ -73,10 +74,9 @@
   cotents wth pretty-print."
   [xml-output-file]
   (let [xml-file (if (instance? File xml-output-file)
-               xml-output-file
-               (.toFile xml-output-file))
-        xml-string (slurp xml-file)
-        pretty-xml-string (pretty-print-xml xml-string)]
+                   xml-output-file
+                   (.toFile xml-output-file))
+        pretty-xml-string (pretty-print-xml xml-file)]
```
alice-dw, line 79: skip the slurp--instead of passing a string to pretty-print-xml, we want the file object itself

src/vip/data_processor/util.clj
```diff
@@ -31,7 +32,10 @@
   "Returns a reader from io/reader, which has advanced past a byte
   order marker if one exists."
   [x & opts]
-  (let [reader (apply io/reader (.toFile x) opts)]
+  (let [file (if (instance? File x)
+               x
+               (.toFile x))
+        reader (apply io/reader file opts)]
```
alice-dw, line 38: same as before, but account for cases where x is already a file by only calling .toFile when necessary

danny-pease (approved):
🌳

dwxander (approved):
🛟

PR #379, opened 2023-03-20
VIP-921: Fix Java outofmemory error

alice-dw:
It was erroring before because we convert the whole xml file to a string, format that string, and write the formatted string back to the file. This fails when the file is too large. Instead, we will read the xml from a stream, output the transformed xml to a stream, and write that stream to a file.


src/vip/data_processor/s3.clj
```diff
@@ -56,8 +56,9 @@
   [xml]
   (let [in (javax.xml.transform.stream.StreamSource.
             (util/bom-safe-reader xml))
-        writer (java.io.StringWriter.)
-        out (javax.xml.transform.stream.StreamResult. writer)
+        outfile (io/file (.getName xml))
+        output-stream (FileOutputStream. outfile)
+        out (javax.xml.transform.stream.StreamResult. output-stream)
```
alice-dw, line 61: we can write the transform result directly to the outfile like this:
```
FileOutputStream outputStream = new FileOutputStream(new File("outputfile.xml"));
transformer.transform(new StreamSource(agent), new StreamResult(outputStream));
```
so we do the clojure version of that, setting the name of the `outfile` to the name of the original xml file passed in.
[source](https://stackoverflow.com/questions/26046608/java-transformer-how-do-you-make-its-result-into-an-outputstream)

src/vip/data_processor/s3.clj
```diff
@@ -67,18 +68,16 @@
...
   cotents wth pretty-print."
   [xml-output-file]
   (let [xml-file (if (instance? File xml-output-file)
                    xml-output-file
-                   (.toFile xml-output-file))
-        pretty-xml-string (pretty-print-xml xml-file)]
-    (spit xml-file pretty-xml-string)
-    xml-file))
+                   (.toFile xml-output-file))]
+    (pretty-print-xml xml-file)))
```
alice-dw, line 80: skip any string stuff--`pretty-print-xml` takes a file and returns a file now

dwxander (approved):
🗃

danny-pease (approved):
📝

PR #380, opened 2023-03-30
VIP-926: Remove _other_idx from locality temp table

alice-dw:
This is causing an error for indiana data (see attached JIRA card). Little to no documentation on why the index was added in the first place ([original PR](https://github.com/votinginfoproject/data-processor/pull/272/files)), and it seems to work locally without this index. So we will remove it for now and see if that causes issues for other feeds.


resources/migrations/20230329172944-01-remove-locality-other-idx.down.sql
```diff
@@ -0,0 +1,166 @@
...
+    into election_administrations
+    using localities.id;
+
+    raise notice '% finishing up...', clock_timestamp();
+
+    locality_paths :=  localities.precincts || polling_locations || hours_open || street_segments || electoral_districts || contests || election_administrations;
+    insert into v5_dashboard.paths_by_locality values(rid, localities.id, locality_paths);
+  end loop;
+end;
+$$ language plpgsql;
```
alice-dw, line 166: this file should be the same as the up file for the most recent migration* that touches the temp table, found [here](https://github.com/votinginfoproject/data-processor/blob/main/resources/migrations/20170725-01-more-specific-index-for-locality-temp-table.up.sql)

\* these are not migrations. these files just overwrite the `populate_locality_table` functionality each time, and there is no modification or "migration" or any data as the name migration would suggest :/

resources/migrations/20230329172944-01-remove-locality-other-idx.up.sql
```diff
@@ -0,0 +1,125 @@
...
+    into election_administrations
+    using localities.id;
+
+    raise notice '% finishing up...', clock_timestamp();
+
+    locality_paths :=  localities.precincts || polling_locations || hours_open || street_segments || electoral_districts || contests || election_administrations;
+    insert into v5_dashboard.paths_by_locality values(rid, localities.id, locality_paths);
+  end loop;
+end;
+$$ language plpgsql;
```
alice-dw, line 125: this file should be the same as above, but without the creation of `_other_idx` on the temp table

alice-dw:
this change works for indiana (zip) & los angeles (xml) data locally. however, i was not able to reproduce the bug locally on main

resources/migrations/20230329172944-01-remove-locality-other-idx.down.sql
danny-pease: thoughts on adding a comment to these "migrations" explaining what they do and offering some context? seems like a place as any to add documentation

danny-pease (approved):
👍

resources/migrations/20230329172944-01-remove-locality-other-idx.up.sql
```diff
@@ -1,3 +1,10 @@
+-- This file (among others) is not a true migration. It replaces 
+-- `v5_dashboard.populate_locality_table`, as the first line suggests. 
+-- Each time we want to make a change to this function, we are not
+-- incrementally applying migrations per the usual pattern, but instead
+-- replacing the functionality of the previous file with the contents 
+-- of the new one.
```
danny-pease, line 6: :guy_fieri_chef_kiss:

dwxander (approved):

PR #381, opened 2023-03-30
VIP-926: Add locality index back in 

alice-dw:
undoes [this change](https://github.com/votinginfoproject/data-processor/pull/380) because the original issue did not reproduce in dev, staging, or prod environments. added new migrations to avoid messiness because the migrations to remove the index are already in staging. but these files are basically the opposite of the previous change (the up is down, down is up) so the final state is reverted to what it was before #380


alice-dw:
[successful run of indiana data in prod](https://dashboard.votinginfoproject.org/#/feeds/2023-05-02-Local-Indiana-17182) 🤪

danny-pease (approved):
🤡

dwxander (approved):
😶‍🌫️

